PR #4, opened 2022-01-12
Use ecbundle from Github and setup build actions

reuterbal:
This changes use of ecbundle to the new Github hosting location.

Also, this sets-up a build workflow using GNU 9.3 via Github actions. This cannot test all variants, though, as we rely on an internally hosted Serialbox version and Loki, which are both not publicly available.

Actual execution of cloudsc is deferred as it should be carried out using runners on appropriate target architectures, eg on the Atos.


mlange05:
Fantastic stuff, and good to go!

PR #5, opened 2022-01-17
Validate results for targets in Github actions

reuterbal:
This adds an additional step in the build workflow to validate results for a small set of columns (100) for double precision binaries.

Notably, they currently show very small deviations on Github which aren't present on leap42 with same compiler version. Small enough though to not be flagged by the validation module. Will investigate this on a separate occasion as it is currently not reproducible locally.


mlange05 (approved):
Great! Deviations look very small indeed. Good to go from me.

PR #7, opened 2022-02-08
Switch to HDF5 as default input format

reuterbal:
As discussed, this switches the default input and reference file library to HDF5. I have tried to clean-up the package discovery process and bundle arguments as good as possible. README should reflect the appropriate changes.

Included is also my small serialbox2hdf5 script that I'm using to convert Serialbox data to HDF5 inputs. That way, everything necessary to update reference data is in one place and since we don't need this in any other context I don't think it would make sense to host this elsewhere.

Also, since Leap42-GNU doesn't support OpenACC I have added `ENABLE_ACC=OFF` to the relevant toolchain file (because I keep forgetting this and there is no reason why this option shouldn't explicitly be disabled that way).


mlange05:
Looks great!. I've pushed a small set of hotfixes for the internal `volta` machine, but otherwise GTG from me.

PR #12, opened 2022-04-21
Test NVHPC in Github actions

reuterbal:
Small addition to our Github actions configuration that verifies builds with NVHPC using the bootstrap script from Atlas.

This can't actually validate results as we can't run the GPU variants but might highlight compile-time problems, in particular with OpenACC annotations, early on.

Takes a bit longer than the other jobs as it needs to download nvhpc and compile hdf5 from source (which takes ~3 and 10 mins, respectively).


mlange05 (approved):
Apologies for the delay in reviewing this. It's looking great and the increased coverage is much appreciated!

PR #6, opened 2022-02-07
[WIP] OpenMP5 GPU version of SCC prototype

dindon-sournois:

PR #15, opened 2022-07-04
OpenMP offload version with CMake integration

reuterbal:
This picks up the code version from PR #6, with @dindon-sournois excellent work on an OpenMP offload variant for CLOUDSC modeled after the SCC-Hoist OpenACC variant. See that PR for some more context. I have added a few commits on top to improve the integration with CMake and ecbundle.

In detail, this adds a new dwarf variant dwarf-cloudsc-gpu-omp-scc-hoist using OpenMP target pragmas and some corresponding data statements in common modules.

We have two variants of that, one using `omp target teams distribute` and the other using the newer `omp target teams loop`. The latter gives better performance on Nvidia but isn't as widely supported, yet, I added some `try_compile` checks that should allow figuring out automatically which of the two variants to build.

Also, I took the opportunity to declutter some of our CMake in the process, hoping to make it a bit more streamlined.

1) Integrate the checks for required packages directly into the ecbuild_add_option calls
2) I made HDF5, OpenMP and OpenACC `PUBLIC` library targets, which should automatically propagate the dependency to all executables that link against this library.

Also note, that `OpenMP_<lang>_FLAGS` variables are overwritten by the `FindOpenMP` module unless they are set as cache variables, which I have fixed for the volta and hpc2020 architecture files. In practice, our user-specified flags have been used anyway during compilation because then the toolchain file gets included again, overwriting the discovered flags from the configuration phase. But with e.g. CCE on AMD, this would fail in the configuration step because some missing OpenMP libraries would not be linked during that stage.

The `FindOpenACC` module has a few issues which I reported to Kitware: https://gitlab.kitware.com/cmake/cmake/-/issues/23691
It looks like the `OpenACC_FOUND` variable issue might be fixed in the next release already.

Note that this might create conflicts with the other open PRs, so feel free to suggest a merge order and I'll rework the others as required.


CMakeLists.txt
```diff
@@ -20,70 +20,61 @@ if( CMAKE_Fortran_COMPILER_ID MATCHES "GNU")
...
-    endif()
-    ecbuild_add_option( FEATURE MPI
-        DESCRIPTION "MPI" DEFAULT OFF
-        CONDITION MPI_Fortran_FOUND )
+ecbuild_add_option( FEATURE MPI
+    DESCRIPTION "MPI" DEFAULT OFF
+    REQUIRED_PACKAGES "MPI COMPONENTS Fortran" )
+if( HAVE_MPI )
+    ecbuild_enable_mpi( COMPONENTS Fortran )
     list(APPEND CLOUDSC_DEFINITIONS HAVE_MPI ${MPI_DEFINITIONS})
```
wdeconinck, line 59: I would not use `ecbuild_enable_mpi` and also not append `${MPI_DEFINITIONS}`
Rather the `MPI::MPI_Fortran` target should be linked where appropriate, if not already done.

arch/toolchains/ecmwf-hpc2020-nvhpc.cmake
```diff
@@ -16,15 +16,19 @@ set( ECBUILD_FIND_MPI ON )
...
 ####################################################################
 
-set( OpenMP_C_FLAGS             "-mp -mp=bind,allcores,numa" )
-set( OpenMP_CXX_FLAGS           "-mp -mp=bind,allcores,numa" )
-set( OpenMP_Fortran_FLAGS       "-mp -mp=bind,allcores,numa" )
+# Note: OpenMP_<lang>_FLAGS gets overwritten by the FindOpenMP module
+# unless its stored as a cache variable
+set( OpenMP_C_FLAGS             "-mp -mp=bind,allcores,numa" CACHE STRING "" FORCE )
+set( OpenMP_CXX_FLAGS           "-mp -mp=bind,allcores,numa" CACHE STRING "" FORCE )
+set( OpenMP_Fortran_FLAGS       "-mp -mp=gpu,bind,allcores,numa" CACHE STRING "" FORCE )
```
wdeconinck, line 23: What if you remove `FORCE`?
If that works, then that would give you option to override via command-line still.

arch/toolchains/ecmwf-hpc2020-nvhpc.cmake
```diff
@@ -16,15 +16,19 @@ set( ECBUILD_FIND_MPI ON )
...
 
 ####################################################################
 # OpenAcc FLAGS
 ####################################################################
 
-set( OpenACC_Fortran_FLAGS "-acc=gpu -gpu=lineinfo,fastmath" )
+# NB: We have to add `-mp` again to avoid undefined symbols during linking
+# (smells like an Nvidia bug)
+set( OpenACC_Fortran_FLAGS "-acc=gpu -mp=gpu -gpu=cc80,lineinfo,fastmath" )
```
wdeconinck, line 31: To be consistent with OpenMP flags we could also add it to the cache and allow to override via command-line.

arch/toolchains/ecmwf-volta-pgi-gpu.cmake
```diff
@@ -16,15 +16,19 @@ set( ECBUILD_FIND_MPI ON )
...
 ####################################################################
 
-set( OpenMP_C_FLAGS             "-mp -mp=bind,allcores,numa" )
-set( OpenMP_CXX_FLAGS           "-mp -mp=bind,allcores,numa" )
-set( OpenMP_Fortran_FLAGS       "-mp -mp=bind,allcores,numa" )
+# Note: OpenMP_<lang>_FLAGS gets overwritten by the FindOpenMP module
+# unless its stored as a cache variable
+set( OpenMP_C_FLAGS             "-mp -mp=bind,allcores,numa" CACHE STRING "" FORCE )
+set( OpenMP_CXX_FLAGS           "-mp -mp=bind,allcores,numa" CACHE STRING "" FORCE )
+set( OpenMP_Fortran_FLAGS       "-mp -mp=gpu,bind,allcores,numa" CACHE STRING "" FORCE )
```
wdeconinck, line 23: Same comments apply here.

arch/toolchains/ecmwf-volta-pgi-gpu.cmake
```diff
@@ -16,15 +16,19 @@ set( ECBUILD_FIND_MPI ON )
...
 
 ####################################################################
 # OpenAcc FLAGS
 ####################################################################
 
-set( OpenACC_Fortran_FLAGS "-acc -ta=tesla:lineinfo,deepcopy,fastmath" )
+# NB: We have to add `-mp` again to avoid undefined symbols during linking
+# (smells like an Nvidia bug)
+set( OpenACC_Fortran_FLAGS "-acc=gpu -mp=gpu -gpu=cc70,fastmath,lineinfo" )
```
wdeconinck, line 31: Same comments apply here.

arch/toolchains/github-ubuntu-gnu.cmake
```diff
@@ -26,3 +25,4 @@ set(ECBUILD_Fortran_FLAGS "${ECBUILD_Fortran_FLAGS} -fno-second-underscore")
 set(ECBUILD_Fortran_FLAGS "${ECBUILD_Fortran_FLAGS} -ffree-form")
```
wdeconinck, line 25: I realise this is not touched in this commit, but is this flags required? If so it is not really a platform-specific flag, but should be embedded in source, i.e. each gfortran compiler would suffer this.

src/cloudsc_c/CMakeLists.txt
```diff
@@ -19,42 +19,39 @@ if( HAVE_CLOUDSC_C )
...
+            cloudsc/cloudsc_validate.h
+            cloudsc/cloudsc_validate.c
+            cloudsc/mycpu.h
+            cloudsc/mycpu.c
+        PUBLIC_INCLUDES
+            $<INSTALL_INTERFACE:include>
+            $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/cloudsc>
+        PUBLIC_LIBS
+            Serialbox::Serialbox_C
+            $<${HAVE_OMP}:OpenMP::OpenMP_C>
```
wdeconinck, line 42: `$<${HAVE_OMP}:OpenMP::OpenMP_C>` is really neat! It had never occured to me!

wdeconinck (changes requested):
Reviewd only the CMake.

arch/toolchains/ecmwf-hpc2020-nvhpc.cmake
reuterbal: Inded, `FORCE` is not required here. However, the situation is even worse and inconsistent to specify custom OpenMP flags:

* For Fortran, the easiest way is to set the variable in the cache, thus keeping the module from overwriting it
* For C/CXX, it requires setting `OpenMP_${LANG}_FLAGS` _and_ `OpenMP_${LANG}_LIB_NAMES` to shortcut the derivation of these two variables entirely, otherwise they get overwritten with `FORCE` by the module.

The latter feels very unnecessary and I might open an issue about it with Kitware.

reuterbal: For demonstration purposes I'll put it in the arch files here with a comment to refer to later on, as the ability to overwrite them may be crucial to certain offloading configurations, where the auto-discovery of flags may just be wrong.

Culprit is [here](https://github.com/Kitware/CMake/blob/c4cdd7a24d3016ccf91fae26646a548b52de4a78/Modules/FindOpenMP.cmake#L475) in the module

arch/toolchains/ecmwf-hpc2020-nvhpc.cmake
reuterbal: Agreed, hadn't done it because it wasn't necessary but makes sense absolutely.

arch/toolchains/github-ubuntu-gnu.cmake
reuterbal: I'll try to find out, good point!

CMakeLists.txt
reuterbal: In fact, we had the same redundant construction for HDF5_Fortran_DEFINITIONS

reuterbal:
Thanks a lot for the review and your comments. I think I addressed them all, so should be good for another look!

arch/toolchains/ecmwf-hpc2020-nvhpc.cmake
reuterbal: right, thanks for the reminder!

wdeconinck (approved):
All good to me, as far as CMake is concerned.

arch/toolchains/ecmwf-hpc2020-nvhpc.cmake
reuterbal: Tracked in https://gitlab.kitware.com/cmake/cmake/-/issues/23743

mlange05 (approved):
Ok, first of all this is a major piece of work and immense progress, and I would like to thank everyone involved for this @dindon-sournois, @reuterbal, @wdeconinck ! The only two comments I have is that we can remove the OpenACC annotations in the OpenMP-offload kernel, and maybe a brief README entry to document the compiler-specific directive usage. Otherwise this is super-impressive work, thanks a bunch!

src/cloudsc_gpu/cloudsc_gpu_omp_scc_hoist_mod.F90
```diff
@@ -0,0 +1,2658 @@
...
+    REAL(KIND=JPRB) :: ZEPSILON
+
+    REAL(KIND=JPRB) :: ZCOND1, ZQP
+
+    REAL(KIND=JPRB) :: PSUM_SOLQA
+
+
+#include "fcttre.func.h"
+#include "fccld.func.h"
+!$acc routine seq
```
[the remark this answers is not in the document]
reuterbal: Agreed, will do

src/cloudsc_gpu/cloudsc_driver_gpu_omp_scc_hoist_mod.F90
```diff
@@ -0,0 +1,236 @@
...
+!$omp map(from: &
+!$omp   pfsqlf,pfsqif,pfcqnng, &
+!$omp   pfcqlng ,pfsqrf,pfsqsf,pfcqrng,pfcqsng,pfsqltur, &
+!$omp   pfsqitur,pfplsl,pfplsn,pfhpsl,pfhpsn)
+
+    ! Local timer for each thread
+    TID = GET_THREAD_NUM()
+    CALL TIMER%THREAD_START(TID)
+
+#ifdef HAVE_OMP_TARGET_LOOP_CONSTRUCT
```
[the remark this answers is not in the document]
reuterbal: It is standard version compliance-specific behaviour. The loop construct was added with OpenMP 5.0 but not all compilers implement it. We switch between the versions automatically by compiling snippets with loop construct and target teams distribute in the CMake configuration. If the first is supported, we use that, otherwise fall back to the second, and switch off this variant completely if neither works.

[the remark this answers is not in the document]
reuterbal: I'll mention it in the README, though

mlange05:
Also, I forgot to mention this earlier: Confirmed functional and fast on two GPU-enabled systems, the in-house "volta" machine and a EuroHPC machine (Meluxina, will contribute arch files soon). :+1:

mlange05:
Ok, I'll merge this now. Any follow-up fixes (docs ,arches, etc.) can be done in a separate PR.

PR #17, opened 2022-08-31
Fix undefined values when gidx is multiple of nlon

reuterbal:
As reported by @lukasm91


mlange05 (approved):
Looks good, and GTG! Many thanks, @lukasm91 for reporting!

PR #21, opened 2022-10-12
GPU: CUDA-Fortran variant SCC-CUF

mlange05:
This brings in the consolidated CUDA-Fortran version created by @marsdeno (many thanks!) that uses the SCC loop layout, but implements device dispatch via CUDA-Fortran (CUF). This new variant can be anbled via bundle option `--with-cuda` and demonstrates how similar performance to SCC-hoist can be achieved without the need to hoist temporary arrays to the driver layer.

A few details and remarks:
* Kernel is based on SCC data layout, but with explicit `device` annotations for local variables and temporary arrays
* `NLEV` vertical dimension is parameterised as `137` and compiled through(!), thus allowing the CUDA compiler to unroll loops and schedule temporaries better
* Two approaches are demonstrated for parameter sets:
  * In a derived type, we move the entire object to device memory in the driver, just before entering the kernel
  * In module scope, parameter variables sets are assigned to constant memory, with large assignment utility routines that are templateable (eg. `YOMCST_UPDATE_DEVICE`); both are valid and `device` vs. `constant` does not cause much performance difference
* Since CMake natively understands CUDA, but not CUDA-FOrtran (`.cuf`) we emulate dealing with dedicated CUDA-Fortran files by explicitly adding `-Mcuda` only to those files, this includes:
  * the additional parameter management routines in `common` 
  * the driver and kernel routines in `cloudsc_gpu`


FussyDuck:
[CLA assistant check](http://claassistant.ecmwf.int:80/ecmwf-ifs/dwarf-p-cloudsc?pullRequest=21) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you all sign our [Contributor License Agreement](http://claassistant.ecmwf.int:80/ecmwf-ifs/dwarf-p-cloudsc?pullRequest=21) before we can accept your contribution.<br/>**1** out of **2** committers have signed the CLA.<br/><br/>:white_check_mark: mlange05<br/>:x: marsdeno<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](http://claassistant.ecmwf.int:80/check/ecmwf-ifs/dwarf-p-cloudsc?pullRequest=21) it.</sub>

src/cloudsc_gpu/nlev_mod.F90
```diff
@@ -0,0 +1,3 @@
+MODULE NLEV_MOD
+  INTEGER, PARAMETER :: NLEV=137
```
reuterbal, line 2: Should we have a sanity check somewhere in the inputs that validates this matches the hard-coded value?

reuterbal (changes requested):
This is fantastic, many thanks! The set of changes is indeed rather compact and mostly evolves around annotations with attributes.

I have confirmed this builds and runs fine on ac using the existing environment file. Performance is even slightly better than SCCH.

Can you add a description of this to the README please?

src/cloudsc_gpu/CMakeLists.txt
```diff
@@ -23,6 +23,11 @@ ecbuild_add_option( FEATURE CLOUDSC_GPU_SCC
     CONDITION Serialbox_FOUND OR HDF5_FOUND
 )
 
+ecbuild_add_option( FEATURE CLOUDSC_GPU_SCC_CUF
+    DESCRIPTION "Build optimized GPU version of CLOUDSC using SCC layout in CUDA" DEFAULT OFF
+    CONDITION Serialbox_FOUND OR HDF5_FOUND
```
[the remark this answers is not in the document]
marsdeno: Yes, should definitely have a `HAVE_CUDA` there

src/common/module/yoethf.cuf.F90
```diff
@@ -0,0 +1,88 @@
...
+
+IMPLICIT NONE
+
+SAVE
+
+!     ------------------------------------------------------------------
+!*     *YOETHF* DERIVED CONSTANTS SPECIFIC TO ECMWF THERMODYNAMICS
+!     ------------------------------------------------------------------
+
+REAL(KIND=JPRB) , CONSTANT :: R2ES_D
```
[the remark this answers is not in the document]
marsdeno: Correct

src/cloudsc_gpu/nlev_mod.F90
marsdeno: I was lazy doing it this way, it would not have been very difficult to pass NLEV down from driver call site.

mlange05: This one might actually go a bit deeper, as I actually tried to remove it and got CUDA runtime errors! Will need to re-create to give more details, but I don't think this is just about performance boost. I could be wrong, but I think the run-time was struggling with two-dimensional automatic arrays, only that increasing the the CUDA_HEAPSIZE did not help this time. I can have another look...

marsdeno: Interesting, pretty sure I had it working without `nlev` as parameter, will look through my dev branch to see if I can find it.

marsdeno (approved):

reuterbal:
Having confirmed that the value for NLEV is indeed required to be a compile-time constant, this is now ready for merging. Thanks again for the fantastic work @marsdeno and @mlange05.

reuterbal (approved):

PR #23, opened 2022-11-16
JUBE benchmark configuration for CLOUDSC

reuterbal:
This brings in the JUBE benchmark configuration for CLOUDSC, simplifying the execution of variants in a standardized fashion on different systems. It is created in a composable way, allowing to override benchmark definitions or configuration values as needed in platform-specific JUBE configurations.

Included is a collection of platform configurations as well as their corresponding arch files. These have been used for the last cross-platform study for the reporting over the summer.


mlange05 (approved):
Fantastic contribution, and very well documented. I've only used bits and pieces of this, but in general it's very well set up. GTG from me.

reuterbal:
Thanks for pointing these out, copy-paste left-overs from older toolchain files. Should be cleaned-up now.

PR #27, opened 2022-12-08
Hotfix: Use 20.04 runner image and make sure GNU 9.4 is installed

reuterbal:
Github is transitioning ubuntu-latest to 22.04 runner images and so we end up with different compiler versions randomly. This should make sure we are using GNU 9.4 every time.


mlange05 (approved):
Uncontroversial, many thanks.

PR #28, opened 2022-12-08
Python/GT4Py: Insert missing copyright headers on source and scripts

mlange05:
Insert the missing copyright headers for the new Python variant (sources and scripts).

### DIFF
--- a/src/cloudsc_python/setup.cfg
+++ b/src/cloudsc_python/setup.cfg
@@ -1,8 +1,8 @@
 [metadata]
 name = cloudsc4py
 description = GT4Py-based implementation of the CLOUDSC dwarf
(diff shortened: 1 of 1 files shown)


stubbiali (approved):
Thanks a lot for taking care of the license statements, @mlange05! All good. You may also want to double-check the package metadata specified in `setup.cfg`. Particularly, only ETH Zurich is listed as author, and I included my email address as main contact. Should we add ECMWF to the list of authors and append another contact (or change it)?

reuterbal:
Many thanks for taking care of this @mlange05 and @stubbiali. GTG!

reuterbal (approved):

PR #30, opened 2022-12-14
Pyinterface - Variant to execute Fortran kernels from Python

piotrows:
This PR adds option to execute CLOUDSC driver from Python driver. The code is tested on Atos (Intel/Gnu) and Mac M1. In addition, there is a fix to mycpu.c and CMakeLists.txt (sequential build of static and dynamic common library) to enable parallel build with cmake -G Ninja && ninja.


FussyDuck:
[CLA assistant check](http://claassistant.ecmwf.int:80/ecmwf-ifs/dwarf-p-cloudsc?pullRequest=30) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](http://claassistant.ecmwf.int:80/ecmwf-ifs/dwarf-p-cloudsc?pullRequest=30) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](http://claassistant.ecmwf.int:80/check/ecmwf-ifs/dwarf-p-cloudsc?pullRequest=30) it.</sub>

src/cloudsc_fortran_pyiface/pythonsrc/cloudsc_data.py
```diff
@@ -0,0 +1,382 @@
...
+from collections import OrderedDict
+import h5py
+import numpy as np
+import cloudsc as clsc
+
+
+NCLV = 5      # number of microphysics variables
+
+
+def define_fortran_fields(nproma,nlev,nblocks):
```
mlange05, line 18: Why not turn this into a class? Instead of storing fields in `locals()` you could attach them to the `CloudscFields` object via `setattr`, if you would like to keep the dynamic introspection loops below? But you'd create these in the constructor and allow the outer layers to access them via 
```
fields = CloudscField(...)
fields.load(...)
fields.prainfrac_toprfz  # This is data object
```

src/cloudsc_fortran_pyiface/pythonsrc/cloudsc_data.py
```diff
@@ -0,0 +1,382 @@
...
+
+
+    with h5py.File(path, 'r') as f:
+        fields['KLON'] = f['KLON'][0]
+        fields['KLEV'] = f['KLEV'][0]
+        fields['PTSPHY'] = f['PTSPHY'][0]
+
+        for argname in argnames_nlev:
+            print('Loading field:',argname)
+            fields[argname] = np.asfortranarray(
```
mlange05, line 127: If it's used this often, this could be a private class method, eg.:
```
class CloudscFields:
...

  def _numpy_from_fortran(arr):
      return np.asfortranarray( np.transpose(np.reshape(...)
```

src/cloudsc_fortran_pyiface/pythonsrc/cloudsc_data.py
```diff
@@ -0,0 +1,382 @@
...
+                                 fields['tendency_tmp_cld'])
+    return fields
+
+def pack_buffer_using_tendencies(buffervar,tendency_a,tendency_t,tendency_q,tendency_cld):
+    """
+    pack_buffer_using_tendencies serves as a packager of a single-variable
+    (that may consist of multiple fields, e.g. moist species)
+    tendencies into a continous buffer
+    """
+    buffervar[:,:,0       ,:]=tendency_t  [:,:,:]
```
mlange05, line 180: Please don't do Fortran index-alignment - any linter will flag that!

src/cloudsc_fortran_pyiface/pythonsrc/cloudsc_data.py
```diff
@@ -0,0 +1,382 @@
...
+
+        for argname in argnames_nlevp:
+            print('Loading field:',argname)
+            fields[argname] = np.asfortranarray(
+                                 np.transpose(
+                                    np.reshape(
+                                       np.ascontiguousarray(f[argname.upper()]),
+                                       (nblocks,nlev+1,nproma),order='C')))
+
+        for argname in argnames_withnclv:
```
mlange05, line 141: I personally think the dynamic inflection here is almost over-engineered. Simply moving the casting into a utility routine, or doing it as part of an "allocate" or "load" helper routine would allow you to simply list the fields here. However, this comes down primarily to taste, so it doesn't need to get addressed here - just a comment.

src/cloudsc_fortran_pyiface/pythonsrc/run_python_driven_cloudsc.sh
```diff
@@ -0,0 +1,8 @@
+module load python3
+python3 -m venv ~/cstest
+source ~/cstest/bin/activate
+pip install --upgrade pip
+pip install f90wrap h5py
```
mlange05, line 5: Again, everything above the export should be dealt with during the install / make phase. For the run phase, I think we can keep a little runner script like this, although I think that the path resolution could also be done in the driver with some additional configuration.

src/cloudsc_fortran_pyiface/CMakeLists.txt
```diff
@@ -0,0 +1,132 @@
...
+   add_custom_target( pythonexecs ALL SOURCES ${PYTHON_SOURCES}
+         COMMAND ${CMAKE_COMMAND} -E copy_directory ${CMAKE_CURRENT_SOURCE_DIR}/pythonsrc  ${CMAKE_CURRENT_BINARY_DIR}/pythonsrc
+         COMMAND ${CMAKE_COMMAND} -E copy_directory ${CMAKE_CURRENT_BINARY_DIR}/pythonsrc  ../../bin/pythonexec
+         COMMAND sh -c "rm -f ../../bin/pythonexec/*.f90"
+         COMMENT "Copy Python driver to Fortran kernel to executable directory"
+        )
+
+    target_link_libraries( ${DWARF_CLOUDSC_LIB} PRIVATE cloudsc-common-lib HDF5::HDF5 )
+    # Define the binary build target for this variant
+    ecbuild_add_executable( TARGET dwarf-cloudsc-fortran-pyiref
```
mlange05, line 68: This target currently fails to build in all CI runners.

src/cloudsc_fortran_pyiface/CMakeLists.txt
```diff
@@ -0,0 +1,132 @@
...
+            ./fortransrc/dwarf_cloudsc.F90
+        DEFINITIONS ${CLOUDSC_DEFINITIONS}
+    )
+
+    target_link_libraries( dwarf-cloudsc-fortran-pyiref PRIVATE cloudsc-common-lib ${DWARF_CLOUDSC_LIB})
+
+    target_include_directories(dwarf-cloudsc-fortran-pyiref PRIVATE ${CMAKE_CURRENT_BINARY_DIR})
+    target_include_directories(dwarf-cloudsc-fortran-pyiref PRIVATE ${CMAKE_CURRENT_BINARY_DIR}/../common)
+
+    if( HAVE_OMP AND TARGET OpenMP::OpenMP_Fortran )
```
mlange05, line 79: Aren't things like HDF5 / OpenMP / MPI / OpenACC linkage inherited from libcommon these days? I don't think this is necessary, or am I missing something here?

src/cloudsc_fortran_pyiface/CMakeLists.txt
```diff
@@ -0,0 +1,132 @@
...
+    # Define the binary build target for this variant
+    ecbuild_add_executable( TARGET dwarf-cloudsc-fortran-pyiref
+        SOURCES
+            ./fortransrc/dwarf_cloudsc.F90
+        DEFINITIONS ${CLOUDSC_DEFINITIONS}
+    )
+
+    target_link_libraries( dwarf-cloudsc-fortran-pyiref PRIVATE cloudsc-common-lib ${DWARF_CLOUDSC_LIB})
+
+    target_include_directories(dwarf-cloudsc-fortran-pyiref PRIVATE ${CMAKE_CURRENT_BINARY_DIR})
```
mlange05, line 76: Why is this needed?

src/cloudsc_fortran_pyiface/pythonsrc/dwarfdriver.py
```diff
@@ -0,0 +1,83 @@
...
+from cloudsc_data import cloudsc_validate
+sys.path.append('../../build/src/cloudsc_fortran_pyiface')
+sys.path.append('../../build/lib')
+sys.path.append('.')
+
+NPROMA=100
+NUMOMP=1
+NLEV=137
+NGPTOT=100
+NGPTOTG=100
```
mlange05, line 24: We should add some CLI interface for the runner here to vary NUMOMP and NPROMA, but that can be left for a follow-on PR. I recomment using [click](https://click.palletsprojects.com/en/8.1.x/).

src/cloudsc_fortran_pyiface/pythonsrc/dwarfdriver.py
```diff
@@ -0,0 +1,83 @@
...
+import sys
+from pathlib import Path
+from operator import itemgetter
+import cloudsc as clsc
+from cloudsc_data import define_fortran_fields
+from cloudsc_data import load_input_fortran_fields, load_input_parameters
+from cloudsc_data import load_reference_fields
+from cloudsc_data import convert_fortran_output_to_python
+from cloudsc_data import cloudsc_validate
+sys.path.append('../../build/src/cloudsc_fortran_pyiface')
```
mlange05, line 16: Since you're already setting relative paths here, I'd move the import of `cloudsc as clsc` further down and possibly force the load-path detection here. Something like (untested):
```
from importlib import import_module

here = Path(__file)
cldir = here / '<relative path to cloudsc parent directory>'

if cldir not in sys.path:
    sys.path.append(cldir)

cloudsc = import_module('cloudsc')

File: src/common/module/fcttre_mod.F90 [hunk 11]
```diff
@@ -31,20 +31,99 @@ MODULE FCTTRE_MOD
 !     ------------------------------------------------------------------
 
   USE PARKIND1, ONLY : JPIM, JPRB
-
-  USE YOMCST, ONLY : RG, RD, RCPD, RETV, RLVTT, RLSTT, RLMLT, RTT, RV, RA, RPI
-  USE YOETHF, ONLY : R2ES, R3LES, R3IES, R4LES, R4IES, R5LES, R5IES, &
+  IMPLICIT NONE
+  REAL(KIND=JPRB),PUBLIC, PROTECTED :: &
+ &                 RG, RD, RCPD, RETV, RLVTT, RLSTT, RLMLT, RTT, RV, RA, RPI,          &
+ &                 R2ES, R3LES, R3IES, R4LES, R4IES, R5LES, R5IES,                     &
```
mlange05, line 37: I'm not 100% convinced that this is / will be neutral for all the other variants. Is there a way to make this an optional mode ? Or should we revisit these in a separate PR to test and evaluate in isolation - this could have quite an impact on all the other variants, if it isn't explicitly protected.

mlange05 (changes requested):
Ok, this change has actually grown quite extensively, and there are still a list of issues that need to be addressed. Overall, I am happy with the approach taken (cmake-f90wrap / parameter object with fuzzed comdeck function / runtime load-and-run via hdf5), but in particular the packaging of the Python require a little work. I've tried to include some helpful examples, but we can take this offline if a more detailed discussion is required.

To move forward with the PR, I recommend splitting this into three pieces:
* Finish PR #30 to get the Mac adaptation out of this
* Create a separate branch with the comdeck changes (fcttre / fccld) and the utility types (yoethf / yomcst). This should create a clean example (possibly in cloudsc_fortran) where we pass all of the parameter types as argument. We can re-visit the mechanism by which this is done here - as we've recently seen an alternative approach to this in a separate project. 
* Once this is clean, adjust the python packaging to introduce `cloudsc_pyinterface` as a pip-installable package, and use cmake to bootstrap the venv (example linked). Then all we need to do is fuzz (and lint!) the cloudsc_data module, to get this clean. Eventual further integration with the other python variant can then follow later.

mlange05:
Ok, I think this has been superseded by PR #38, so closing this PR.

PR #35, opened 2023-01-18
Remove dependence on the global modules from the cloudsc-fortran

piotrows:
This commit removes the dependence on the parameters from global modules from the clouds computation.


src/cloudsc_fortran/dwarf_cloudsc.F90
```diff
@@ -87,13 +103,143 @@ PROGRAM DWARF_CLOUDSC
...
+ & RALFDCP, RKOOP1, RKOOP2
+USE YOECLDP    , ONLY : YRECLDP
+USE YOEPHLI    , ONLY : YREPHLI
+
+TYPE(TOMCST)      ,INTENT(INOUT) :: YDOMCST
+TYPE(TOETHF)      ,INTENT(INOUT) :: YDOETHF
+TYPE(TECLDP)      ,INTENT(INOUT) :: YDECLDP
+TYPE(TEPHLI)      ,INTENT(INOUT) :: YDEPHLI
+
+YDECLDP%RKCONV          = YRECLDP%RKCONV
```
mlange05, line 135: Should these loader routines not live in the respective modules, like YOETHF, YOEMCST, etc.? We already have global loader functions in there, and we could probably add type-specific ones there (either `YDETHF%INITIALIZE()` or add something like `YOETHF_LOAD_PARAMETERS(YDETHF)` to the existing loader routine? This would avoid polluting the high-level entry routines with this.

src/cloudsc_fortran/dwarf_cloudsc.F90
```diff
@@ -68,8 +81,11 @@ PROGRAM DWARF_CLOUDSC
 ! TODO: Create a global global memory state from serialized input data
 CALL GLOBAL_STATE%LOAD(NPROMA, NGPTOT, NGPTOTG)
 
+CALL INITIALIZE_PARAMETER_DATATYPES(YDOMCST, YDOETHF, YDECLDP, YDEPHLI )
+
 ! Call the driver to perform the parallel loop over our kernel
-CALL CLOUDSC_DRIVER(NUMOMP, NPROMA, GLOBAL_STATE%KLEV, NGPTOT, NGPTOTG, &
+CALL CLOUDSC_DRIVER( NUMOMP, NPROMA, GLOBAL_STATE%KLEV, NGPTOT, NGPTOTG, &
+     & NCLDQV, NCLDQL, NCLDQR, NCLDQI, NCLDQS, NCLV,  &
```
mlange05, line 88: The global cloud vector indexes/sizes can probably also be wrapped in a derived type to avoid passing all of them down.

mlange05 (changes requested):
Overall looks good, and gets us around a very big headache problem, and is in line with developments brought to us from Meteo-France. 

I've marked to small cosmetic changes, one about the location of the additional loader code (not in the dwarf head program), and the additional integer flags that could also be wrapped, rather than extending the subroutine signature. After that, this should be GTG.

src/cloudsc_fortran/dwarf_cloudsc.F90
piotrows: Is derived type necessary or it could be just a vector ? Perhaps it could be hardcoded (e.g. called integer :: TNCLV(NCLV)) in MODULE YOECLDP ?

mlange05: These are named integer indexes. Why would we want to create yet another pattern, instead of using the same as for the other module-level parameters? Passing them as a derived type object we can use associates in the subroutine body, following the existing convection, no?

piotrows: @mlange I understand that explicit declaration of NCLDQV, NCLDQL, NCLDQR, NCLDQI, NCLDQS, NCLV must stay as they are in the yoecldp, without putting them in the derived type envelope. So the information needs to  be duplicated somewhere, what about putting them as members of TECLDP, so no extra type definition is need ?

piotrows: Moreover, in the current form they can be resolved at compile time (since these are parameters). If I include them as variables in the derived type, then it won't be possible anymore. So repeating them in the derived type as parameters seems to be desirable option.

mlange05: Yes, adding them to TECLDP and setting them in the initialisation routine should work nicely I think.

mlange05:
**[DO NOT MERGE UNTIL RESOLVED!]** Not quite sure what's going on here, but this change seems to drop CPU performance of the changed variant significantly; I'm getting a drop from around 88GF/s to 58GF/s! Please hold off on review/merge until this is understood.

piotrows:
> Intel

piotrows:
I have checked that the performance on Mac+gfortran 12 doesn't depend significantly on the availability of cloud indices 1...5 at compile time. To be checked on ATOS.

piotrows:
I have checked that with Intel the current commit has matching performance to the 'Apple' commit. It seems that with gcc the performance is less dependent on the strategy of passing of indices.

mlange05 (approved):
Thank you. With the latest modifications the change looks good and ready to go. I've verified performance impact is neutral on Atos/Intel.

PR #38, opened 2023-02-08
PyIface - Fortran-Python bridging with CMake compilation

piotrows:
Working example of packaged pyiface driver that may be invoked e.g. by ctest on ATOS. CMake side may seem polished, Python side not fully.


src/cloudsc_pyiface/src/cloudscpytools/cloudsc_data.py
```diff
@@ -0,0 +1,457 @@
...
+                                      np.zeros(shape=dims, dtype="float64")))
+            clsc.expand_mod.expand_r1(lfield, b2field,  nlon, nproma, ngptot, nblocks)
+            bfield=b2field
+        elif ldims == 3:
+            b3field=np.asfortranarray(np.transpose(np.zeros(shape=dims, dtype="float64")))
+            clsc.expand_mod.expand_r2(lfield, b3field,  nlon, nproma, nlev, ngptot, nblocks)
+            bfield=b3field
+        elif ldims == 4:
+            b4field=np.asfortranarray(np.transpose(np.zeros(shape=dims, dtype="float64")))
+#Note that we are using expand_r3bis here.
```
mlange05, line 127: Please indent comments the same way as code.

src/cloudsc_pyiface/src/cloudscpytools/cloudsc_data.py
```diff
@@ -0,0 +1,457 @@
...
+#Original expand_r3 does not pass nlev from python to Fortran
+#            print ("This is 4D field")
```
mlange05, line 129: Debug leftover?

src/cloudsc_pyiface/src/cloudscpytools/cloudsc_data.py
```diff
@@ -0,0 +1,457 @@
...
+        fields['KLEV'] = f['KLEV'][0]
+        fields['PTSPHY'] = f['PTSPHY'][0]
+        klon=fields['KLON']
+
+        for argname in argnames_nlev:
+            print('Loading field:',argname)
+            fields[argname] = field_c_to_fortran((nblocks,nlev,nproma),
+                                                 f[argname.upper()],nparms,klon)
+
+        for argname in argnames_nlevp:
```
mlange05, line 205: Grouping and iterating seems overkill here. Why not simply list them as 
```
fields['pt'] = ield_c_to_fortran(('PT', dims=(nblocks,nlev,nproma), ...)
```

src/cloudsc_pyiface/src/cloudscpytools/cloudsc_data.py
```diff
@@ -0,0 +1,457 @@
...
+            print('Loading field:',argname)
+            fields[argname] = field_c_to_fortran((nblocks,nlev+1,nproma),
```
mlange05, line 207: If memory serves, hdf5 presents fields as numpy arrays, so there is no need to manage the dimensions explicitly. Instead we could simply interrogate `f[argname].shape` here and manipulate it accordingly, no?

src/cloudsc_pyiface/src/cloudscpytools/cloudsc_data.py
```diff
@@ -0,0 +1,457 @@
...
+    """
+    load_input_parameters returns:
+    - four parameter datatypes that are filled using names read from the reference .h5 file
+    """
+    with h5py.File(path, 'r') as f:
+        tecldp_keys = [k for k in f.keys() if 'YRECLDP' in k]
+        for k in tecldp_keys:
+            attrkey = k.replace('YRECLDP_', '').lower()
+            setattr(yrecldp, attrkey, f[k][0])
+        setattr(yrecldp, 'ncldql', 1)
```
mlange05, line 267: What's wrong with simply `yrecldp.ncldqp = 1`?

src/common/module/yoecldp.F90
```diff
@@ -366,6 +372,11 @@ SUBROUTINE YRECLDP_LOAD_PARAMETERS()
     ! The last two are actually arrays, so treat them as fields
     CALL LOAD_ARRAY('YRECLDP_RBETA', 1, 101, 101, 101, YRECLDP%RBETA(0:100))
     CALL LOAD_ARRAY('YRECLDP_RBETAP1', 1, 101, 101, 101, YRECLDP%RBETAP1(0:100))
+    YRECLDP%NCLDQL = NCLDQL
```
mlange05, line 375: Not too keen on this tbh. We know that actually using these from the Fortran side will make the code slow, so there's no obvious gain here, but increased risk someone's going to fall for that.

mlange05 (commented):
Ok, first of all thank you for a very comprehensive effort and all the work that went into this PR. I think conceptually this is a great starting point, as it allows us to use the cmake layer for the backend compilation, which means we can inherit all the compile settings for the Fortran-side natively. It is not yet clear to me, why your runs are slower compared to the baseline, but we can investigate this separately.

For the PR itself, I've left a list of comments, where I think the packaging can be improved. In particular the cmake layer can possibly be a bit more modular, and the combination of pip installs during compile with an attempt at dynamic loading at runtime needs to be untangled. The Python utility routines also have some peculiar styling choices, and can probably be simplified. This also highlights that, in some future work, we can probably consolidate this with the existing `cloudsc_python` package to reduce replication. 

For this PR, I think I can take over from here, since I want to play with this a little, and see if I can simplify a few things. So as a next step I'll play with this and see if I can address some of the comments from the list.

CMakeLists.txt
```diff
@@ -92,6 +92,13 @@ if( HAVE_SERIALBOX )
 endif()
 
 ecbuild_find_package( NAME loki )
+ecbuild_find_python(VERSION 3.8 NO_LIBS)
+
+if ( PYTHON_EXECUTABLE )
```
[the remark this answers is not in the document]
piotrows: Agreed, this is on the TODOs list, couldn't  come up with a better idea for the moment, need a discussion/advice here.

arch/ecmwf/hpc2020/gnu/11.2.0/env.sh
```diff
@@ -35,7 +35,7 @@ module_load hpcx-openmpi/2.10.0
 module_load boost/1.71.0
 module_load hdf5/1.10.6
 module_load cmake/3.20.2
-module_load python3/3.8.8-01
+module_load python3/3.10.10-01
```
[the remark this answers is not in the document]
piotrows: Sticking to a single version of Python on a given machine was a real pain to me through the process, this was just to make things a bit simpler. 3.10 becomes standard on ATOS in May, anyway. If I recall correctly, 3.6 was not cooperating, 3.8 should be OK.

src/cloudsc_pyiface/CMakeLists.txt
```diff
@@ -0,0 +1,152 @@
...
+    DESCRIPTION "Build the Python interface to the (slightly customized) cloudsc-fortran" DEFAULT ON
+    CONDITION HDF5_FOUND
+)
+ecbuild_add_option( FEATURE CLOUDSC_FORTRAN_PYIFACE_BINARY
+	DESCRIPTION "Build the debug-oriented binary for Python interface of (customized) cloudsc-fortran" DEFAULT OFF
+    CONDITION HAVE_CLOUDSC_FORTRAN_PYIFACE 
+)
+
+if( HAVE_CLOUDSC_FORTRAN_PYIFACE )
+   set (CMAKE_EXPORT_COMPILE_COMMANDS ON)
```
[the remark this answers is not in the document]
piotrows: Debug artefact, sorry.

src/cloudsc_pyiface/src/cloudscpytools/cloudsc_data.py
piotrows: No, rather an urgent note to discuss, because we don't want to carry on with expand_r3bis unless absolutely necessary. A very strange thing, on the Python side nlev was 137, on the Fortran side it was 5.

src/cloudsc_pyiface/src/cloudscpytools/cloudsc_data.py
piotrows: This solution seems much cleaner and portable to me. I am strongly in favour of my solution over the explicit list at the moment.

src/cloudsc_pyiface/src/cloudscpytools/cloudsc_data.py
piotrows: Possibly. On the other hand, the explicit list of dimensions annotates the code here. When playing on the two pianos  at  one time, it seems helpful to have more annotations in the code, at least it was for me. Open to your ideas here.

arch/ecmwf/hpc2020/gnu/11.2.0/env.sh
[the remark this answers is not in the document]
mlange05: True, but if it's not strictly required for this PR, this change should be filed, discussed and tested separately, to avoid knock-on effects with other Python variants.

src/cloudsc_pyiface/src/cloudscpytools/cloudsc_data.py
mlange05: Not sure what "portable" in this context means, but no matter. I can see your point if there's only one set, but of you have to group by dimension, I actually find this unnecessarily hard to read, due to the indirection forcing me to look in different places to find out which shape a field has. And inferring the shape (a tuple) from a naming convention is just plain annoying, but I guess this comes down to taste...

CMakeLists.txt
```diff
@@ -92,6 +92,13 @@ if( HAVE_SERIALBOX )
...
+  execute_process(COMMAND ${PYTHON_EXECUTABLE} --version)
+  execute_process(COMMAND ${PYTHON_EXECUTABLE} -m pip install --upgrade pip)
+  execute_process(COMMAND ${PYTHON_EXECUTABLE} -m pip install f90wrap h5py)
```
mlange05, line 100: We should really pip install the utility Python package here, and let pip figure out that `f90wrap` and `h5py` are the downstream dependencies.

piotrows: But to create the package one needs to have f90wrap already installed ... unless one creates a stub package with dependencies first, perhaps, and then progress with a real one ?

mlange05: Yes, but that is the point of having pip: if your `cloudsc_pyiface` Python package (should really use consistent naming!) specifies these as dependencies, then pip will install all dependencies first before installing the package itself. It will also detect if they are already installed and attempt proper dependency resolution if there are complex dependency chains or if things already exist in a given venv.

piotrows: I  get it, but the pip install can be executed on the package that already exists. Before f90wrap/f2py is executed on the Fortran pieces, the thing we want to install doesn't exist, correct ?

mlange05: It does though. I'm trying to install the data utilities and all the boilerplate stuff, so that we can then get rid of the second pip install at make time. Pip and the Python package dependency resolution should only be executed during configure. You have dynamic module loading in the driver already, so as long as we make sure that that is guarded enough, there is no need to force the second pip install.

mlange05:
Ok, @piotrows , thanks for reviewing and merging this. I'll have another look at this PR once CI is back. In the meantime, I think there are two key things that we should still do for this PR:
* Disable the PyIface variant by default (as discussed in PR #47 )
* Once we merge this use a squash-merge, as the commit list has gotten quite long indeed now and has a lot of back-and-forth

In addition, it'd be great if @reuterbal could take a brief look at this, before final approval?

piotrows:
> 
> * Disable the PyIface variant by default (as discussed in PR [PyIface - Package clean up and refactoring #47](https://github.com/ecmwf-ifs/dwarf-p-cloudsc/pull/47) )
It makes indeed sense to do it, albeit probably at a very last step. So waiting for a @reuterbal comments to come in his "spare time"

reuterbal (changes requested):
Many thanks! It's clear that this has been a major piece of work and I'm already very happy with the current state of this - despite the large number of remarks that I left in the code. Most of them are related to installing the driver script alongside the rest of the package to get around the python venv loading and all of that. As this can become a rats tail easily, I have taken care of those changes in #48 and filed them against this branch, so see if you like them and merge the PR, as it will take care of a large number of the comments already.

Most of all, I think improving the README would be very useful, as well as mentioning it in the repository root's README and linking to your file from there.

src/cloudsc_pyiface/LICENSE
```diff
@@ -0,0 +1,190 @@
+                                 Apache License
```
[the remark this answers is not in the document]
mlange05: We had already linked the LICENSE file in the root, but latest pip does not allow any files outside of the "package directory" to be included in the install manifest, so I simply replicated it here. Could be that this is a `setup.cfg` vs. `pyproject.toml`, but I've not dug deep enough to understand the finer details of this.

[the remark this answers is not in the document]
reuterbal: Oh, interesting, didn't think of that. In that case absolutely fine with this.

.github/scripts/run-targets.sh
```diff
@@ -6,7 +6,7 @@ set -x
 non_mpi_targets=(dwarf-P-cloudMicrophysics-IFSScheme dwarf-cloudsc-c)
 
 # These targets currently cause issues and are therefore not tested
-skipped_targets=(dwarf-cloudsc-gpu-claw)
+skipped_targets=(dwarf-cloudsc-gpu-claw dwarf-cloudsc-fortran-pyiface)
```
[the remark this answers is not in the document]
piotrows: Historically, the showstopper was that the code didn't compile with nvhpc and also couldn't easily deal with venv. Once it is set up for gfortran only in the GitHub test environment, I see no reason why not to test it. Is your changes sufficient to enable testing already ?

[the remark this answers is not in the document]
reuterbal: Yes, see #48 where this is already active

src/cloudsc_pyiface/fortransrc/cloudsc.F90
```diff
@@ -0,0 +1,2905 @@
...
+! In applying this licence, ECMWF does not waive the privileges and immunities
+! granted to it by virtue of its status as an intergovernmental organisation
+! nor does it submit to any jurisdiction.
+
+SUBROUTINE CLOUDSC &
+ !---input
+ & (KIDIA,    KFDIA,    KLON,    KLEV, NCLV, &
+ & PTSPHY,&
+ & PT, PQ, &
+ & PTMPTENT,PTMPTENA,PTMPTENQ,PTMPTENCLD, &
```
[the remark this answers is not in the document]
piotrows: Done

src/cloudsc_pyiface/CMakeLists.txt
```diff
@@ -0,0 +1,158 @@
...
+    # Set up a custom venv for this variant and install the necessary dependencies
+    set( loki_VENV_PATH ${CMAKE_BINARY_DIR}/venv_pyiface )
+    setup_python_venv( ${loki_VENV_PATH} )
+
+    if( NOT Python3_EXECUTABLE )
+        ecbuild_error("[PyIface] Could not find Python3 executable in virtualenv")
+    endif()
+
+    # Install the f90wrap build dependency and via latest pip
+    execute_process(COMMAND ${Python3_EXECUTABLE} -m pip install --upgrade pip)
```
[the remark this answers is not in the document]
mlange05: Yes. The reason is that the actual wrapper-generation/build process is unfortunately not cleanly dependency-driven, but instead a set of commands executed in sequence whenever that lib-object is built. Originally, that even included the venv-creation and install step, but it would mean that all this setup stuff gets repeated every time the lib-target was built, whereas it now only re-does wrapper-compile steps upon re-build. 

This can for sure be fixed via better `add_custom_command` that actually captures the respective dependencies, but I think that can be done in a follow-up step. As a first step for this, we would need to define a proper binary target, instead of dangling everything from a single lib-target via `POST_BUILD`, which ignores dependencies, AFAICT.

FWIW, I'm also not sure if the explicit pip upgrade is still needed, but again, happy to keep this for now and clean up later.

src/cloudsc_pyiface/README.md
```diff
@@ -0,0 +1,42 @@
...
+Similar results can be achieved using GNU compilers on ATOS using module pg and --arch=./arch/ecmwf/hpc2020/gnu/11.2.0/
+
+Troubleshooting:
+If the Python script can't find the newly created Python module containing Fortran code, it is very likely that different Python versions were involved during build and execution. Check if correct python module is loaded and if the venv is clean.
+
+TODOs:
+- the performance is inferior as compared to the reference ?
+- should Python be touched in the topmost cmakelists.txt ? I think current solution is suboptimal. 
+- why ecbuild_find_python can't find Python libs on ATOS ? Non-critical but surprising.
+- find out why expand_r3 does not pass nlev parameter correctly and so expand_r3bis needs to be used
```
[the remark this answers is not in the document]
piotrows: Deleted

src/cloudsc_pyiface/README.md
```diff
@@ -0,0 +1,42 @@
...
+cd build
+. venv_pyiface/bin/activate
+
+# To test performance, execute:
+python ./bin/cloudsc_pyiface.py --numomp=$OMP_NUM_THREADS --ngptot=163840 --nproma=32
+```
+An additional CLI option ``--cloudsc-path=<path-to-python-wrapper>``
+can be used if the build location used to run f90wrap has changed.
+
+Currently, the performance on ATOS is about about 64400 Mflops/s, which is inferior to the reference result of:
```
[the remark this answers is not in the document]
piotrows: done

README.md
```diff
@@ -88,6 +88,14 @@ Balthasar Reuter (balthasar.reuter@ecmwf.int)
...
   manage temporary arrays and needs a large `NV_ACC_CUDA_HEAPSIZE` 
   (eg. `NV_ACC_CUDA_HEAPSIZE=8GB` for 160K columns.)
+- **cloudsc-pyiface.py**: a combination of the cloudsc/cloudsc-driver routines
+  of cloudsc-fortran with the uppermost `dwarf` program replaced with a
+  corresponding Python script capable of HDF5 data load and 
+  verification of computation results. The computation is realized by the
+  Fortran subprogram, mimicking cloudsc-fortran and equipped with only 
+  minor modifications (i.e. derived types/global paramters handling).
+  Turned off by default, activate at the build stage with 
+  '--cloudsc-fortran-pyiface CLOUDSC_FORTRAN_PYIFACE'
```
reuterbal, line 98: This should probably be `--cloudsc-fortran-pyiface=ON`.

README.md
```diff
@@ -339,6 +347,18 @@ The following Loki modes are included in the dwarf, each with a bespoke demonstr
...
 `--with-claw` needs to be specified explicitly.
 
+## Python-driven CLOUDSC variants
+The following partly or fully Python-based CLOUDSC are available:
+- **cloudsc-python**: GT4PY based Python-only implementation. Refer to `src/cloudsc_python`
+  for information on how to bootstrap/execute this variant
+- **cloudsc-pyiface**: Fortran-based CLOUDSC variant driven by the Python script.
+  Activate with: 
+```sh
+./cloudsc-bundle build build --clean --cloudsc-fortran-pyiface=ON
```
reuterbal, line 357: There's a `build` too many.

src/cloudsc_pyiface/CMakeLists.txt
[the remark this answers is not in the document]
reuterbal: Thanks, makes sense. I think we don't need to over-engineer it for CLOUDSC, so fine with this, but may be of interest to further refine the wrapper recipe, of course. All of this is doesn't need to be addressed here, though.

reuterbal (commented):
Thanks for making the changes. There are a few outstanding comments that we should take care of but I think this is almost ready!

src/cloudsc_pyiface/fortransrc/cloudsc.F90
```diff
@@ -0,0 +1,2905 @@
...
+#include "abor1.intfb.h"
+
+!DIR$ VFUNCTION EXPHF
+#include "fcttre.ycst.h"
+#include "fccld.ydthf.h"
+
+!===============================================================================
+!IF (LHOOK) CALL DR_HOOK('CLOUDSC',0,ZHOOK_HANDLE)
+ASSOCIATE( & 
+ & NCLDQV=>YDECLDP%NCLDQV, NCLDQL=>YDECLDP%NCLDQL, NCLDQR=>YDECLDP%NCLDQR, &
```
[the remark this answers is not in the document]
piotrows: Done via new .h file in common/include.

src/common/module/yoecldp.F90
```diff
@@ -215,6 +215,12 @@ MODULE YOECLDP
 INTEGER(KIND=JPIM) :: NAECLBC, NAECLDU, NAECLOM, NAECLSS, NAECLSU
 INTEGER(KIND=JPIM) :: NCLDDIAG
 
+INTEGER(KIND=JPIM) :: NCLDQL   
+INTEGER(KIND=JPIM) :: NCLDQI  
+INTEGER(KIND=JPIM) :: NCLDQR 
+INTEGER(KIND=JPIM) :: NCLDQS    
+INTEGER(KIND=JPIM) :: NCLDQV
```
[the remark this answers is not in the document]
piotrows: Removed.

src/cloudsc_pyiface/fortransrc/cloudsc.F90
```diff
@@ -505,11 +505,11 @@ SUBROUTINE CLOUDSC &
 #include "fcttre.ycst.h"
 #include "fccld.ydthf.h"
 
+#include "micvarsnumbs.h"
```
reuterbal, line 508: I'm not sure this is what @mlange05 had in mind. `NCLDQV`, `NCLDQL`, `NCLDQR`, `NCLDQI`, `NCLDQS` are already declared as parameters in yoecldp.F90, why can't they be imported directly, in the same way as JPRB & co?

mlange05: Yeah, I fail to see how moving them to a new .h file solves any of this, as the Python is not touching this either. In any case, cleanly using existing modules works just as well (I confirmed locally that there's no performance impact), so I'll revert to using only the common module for the constants, and the performance impact analysis can be done in a follow-up.

mlange05:
Ok, I just pushed a last few touch-ups and I think we've addressed all outstanding issues now. In any case, to close this saga, I propose merging this now, as soon as CI is confirmed green, and picking up any other fixes in future PRs and follow-up work. @reuterbal, if you agree, please change your review status to unlock the big green button, and I'll squash-merge this.

reuterbal:
Thanks, Looks great now!

reuterbal (approved):

src/cloudsc_pyiface/fortransrc/cloudsc.F90
piotrows: If I understand correctly, including .h means that the parameter live in the same program unit as the code that uses them, whereas when importing from a module means importing from a "common" library. Overall, however, no objections.

PR #39, opened 2023-02-14
Bug-fix for Loki-SCC/H array demotion

mlange05:
This bug fix adds the header information for yoecldp.F90 to the Loki-SCC and Loki-SCC-H variants. Without this header info, the Loki recipes cannot infer that NCLV=5 and thus arrays with this dimension can be demoted. As a result, both auto-generated Loki-SCC/H variants would compile and run fine, but yield significantly lower performance than their hard-committed counter-parts.

Note: This must have been broken for a while, but I see no way to test this, as the transformation recipes effectively works fine, but is lacking a vital piece of information. Any suggestions for closer monitoring are therefore welcome.


reuterbal (approved):
Nice find! Thanks for the fix.

In terms of finding this, I can only think of two (not mutually exclusive) ways of checking this.
1. We can equip Loki transformations with more debug output where arrays are reported that can't be demoted (essentially an optimization report). This would require manual checking, though.
2. With dedicated runners, e.g. on our HPC, we could set up proper performance testing CI and insert a check that validates sustained performance against an expected minimum.

PR #47, opened 2023-03-20
PyIface - Package clean up and refactoring

mlange05:
_Note: I'm filing this as a separate PR to change key elements of PR #38, as some of the style-changes are fairly invasive._

This set of changes aim to streamline and clarify the underlying components of the "PyIface" variant and thus rectify some of the issues flagged in the original code review. It also also adapts the coding style of the Python component slightly to align it closer to a common "look-and-feel". 

In detail the changes are:
* Stick to common Python version 3.8 and use the provided CMake macros to install the venv during configure
* Refactor the CMake file to separate the three distinct stages of the "PyIface" process: `f90wrap` Fortran-wrapper generation / `f2py-f90wrap` compilation and Py-wrapper generation / custom install of driver script
* Rename Python package (`cloudscpytools` => `pyiface` for consistency) and fix pip install by removing `setup.cfg` and duplicating LICENSE file
  * Note: This works, although a combination of `setup.cfg` / `pyproject.yml` with less replication could also work; I'm open to suggestions.
 * Removed the redundant secondary `pip install` during build, and instead fixed path handling of the dynamic loading of the generated module. For this I also added a manual CLI override, should alternative install locations be needed
 * Added a dedicated `load_module` function to perform dynamic loading in the Python layer, and pass the resulting object as an argument to the data loading routines that require type information
 * Instead of handing a dict of array dimension information down to the data loading layer, we use customary `**kwargs` notation to selectively pull the requested arguments out of the keyword-args dict. (Note, I personally feel that this is a little overkill, and we could simply use explicit keyword-arg passing in each layer, but this becomes quite verbose as some point, so I went for the less invasive option).
* Fixed the bug that necessitated `expand_r3bis` (`ndim` was pulled out of the dims tuple the wrong way), and removed the now redundant Fortran utility routine.
* Updated the Python driver install location and according README entries to reflect the workflow; please test! 

Final note: This PR is intended to address the most glaring issues I flagged in PR #38. However, it does mean that merging this will automatically imply acceptance of #38. Instead I propose a fresh round of review (possibly including @reuterbal as an additional pair of eyes? ) to go over the final adaptation and massaging once this is merged.


reuterbal:
Haven't looked at this PR nor #38 in detail, yet - once this PR is merged I'll happily make up for this!

src/cloudsc_pyiface/drivers/cloudsc_pyiface.py
piotrows, line 1: Use git mv to facilitate diff between old cloudsc_pydriver and new cloudsc_pyiface. We could consider writing cloudsc_py_iface (cloudsc_fortran)  consequently through this dwarf version, pyiface could be too cryptic.

src/cloudsc_pyiface/pyproject.toml
```diff
@@ -23,10 +23,10 @@ classifiers = [
    " Topic :: Scientific/Engineering :: Atmospheric Science "
 ]
 dependencies = [
+    "numpy",
```
piotrows, line 26: I believe we should let f90wrap choose its own numpy version and thus not specify  numpy here. Newest lumpy doesn't support f90wrap AFAIK.

src/cloudsc_pyiface/src/pyiface/__init__.py
```diff
@@ -8,6 +8,5 @@
 # granted to it by virtue of its status as an intergovernmental organisation
 # nor does it submit to any jurisdiction.
 
-
-
-
+from pyiface.cloudsc_data import *  # noqa
+from pyiface.dynload import *  # noqa
```
piotrows, line 12: So should init.py stay empty, or it shouldn't ? I am confused with what I read in the net.

src/cloudsc_pyiface/drivers/cloudsc_pyiface.py
piotrows, line 1: I don't get why there is no default to click cloudsc-path

src/cloudsc_pyiface/CMakeLists.txt
piotrows, line 1: ctest no longer executes our python driver correctly on Mac OS

piotrows, line 1: Appearance of build directory in ./src/cloudsc_pyiface is disturbing, e.g. when dir to dir diff is used.

piotrows (changes requested):
I like the simplifications introduced, however, several functionality regressions should be addressed.

src/cloudsc_pyiface/drivers/cloudsc_pyiface.py
piotrows, line 1: Hard path dependence

src/cloudsc_pyiface/pyproject.toml
mlange05: We do not prescribe a version for exactly this reason, thus leaving it to pip to find an appropriate version that satisfies the respective constraints. However, our own utility routines have an explicit dependency on numpy (`cloudsc_data.py` does `import numpy as np`) and this version free dependency should be explicitly stated.

src/cloudsc_pyiface/drivers/cloudsc_pyiface.py
mlange05: Really now sure what you mean here. Could you please explain in some more detail?

mlange05: Leaving a default option `None` is a customary way to rely on defaults. It's the equivalent of omitting an optional argument, and allows the lowest routine in the argument passing chain to specify the final default, rather than having to propagate the default value (`Path.cwd()`)  explicitly through multiple layers of argument passing.

mlange05: Assuming you mean the `rootpath` option that allows us to find input/reference data? (Please be concise in PR reviews). Agreed, I had meant to fix that - will push a fix.

src/cloudsc_pyiface/CMakeLists.txt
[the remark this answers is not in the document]
mlange05: Not sure what you mean here. Please use inline comments to point out spelling mistakes

mlange05: Yeah, that's probably an artefact of doing a full package install. I believe using "editable" mode will remove this and make things nicer to debug overall. Is this what you're advocating for?

mlange05: Ohhh, that's possible. I'll have a look.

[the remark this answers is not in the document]
piotrows: This line:     ecbuild_error("[PyIface] Could not find Python3 exectubale in virtualenv"). Sorry, must have clicked wrong on GitHub.

piotrows: I am not sure about the exact effect of -e, but I think that adding anything to the src directory in the build process should be avoided. So yes, it sounds right.

src/cloudsc_pyiface/drivers/cloudsc_pyiface.py
piotrows: You have changed the name of the cloudsc_pydriver.py to cloudsc_pyiface.py with mv, not git mv command. Then running diff with git results only in information that one file is deleted and the other is completely new one, making it impossible to compare with git diff. I suppose it is sufficient to rename it back, commit, then use git mv to allow git to trace the change.

src/cloudsc_pyiface/CMakeLists.txt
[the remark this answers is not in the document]
mlange05: :+1:

src/cloudsc_pyiface/src/pyiface/__init__.py
piotrows: Got it.

src/cloudsc_pyiface/drivers/cloudsc_pyiface.py
piotrows: Fine to me.

piotrows: Yes, we have talked about it, so it was just a brief note, perhaps too brief. Sorry.

mlange05:
Ok, I think I've address most comments. @piotrows please have another look.

src/cloudsc_pyiface/drivers/cloudsc_pyiface.py
piotrows: When I go to "Files changed" on GitHub, it still says: 
"
Comment on this file
Load diff
This file was deleted."
Perhaps this is a GitHub thing, then.

piotrows, line 1: I am not sure what is happening, but running test on Mac WHEN ADDED --input-path OPTION, i.e.:
```5: Test command: /Users/piotrows/dwarf-p-cloudsc-naml/build/venv_pyiface/bin/python "cloudsc_pyiface.py" "--numomp=1" "--ngptot=100" "--nproma=16" "--cloudsc-path=/Users/piotrows/dwarf-p-cloudsc-naml/build" "--input-path=/Users/piotrows/dwarf-p-cloudsc-naml/build"```
I am getting:
: Working Directory: /Users/piotrows/dwarf-p-cloudsc-naml/build/bin
5: Environment variables:
5:  OMP_NUM_THREADS=1
5: Test timeout computed to be: 1500
5: Traceback (most recent call last):
5:   File "/Users/piotrows/dwarf-p-cloudsc-naml/build/bin/cloudsc_pyiface.py", line 141, in <module>
5:     main()
5:   File "/Users/piotrows/dwarf-p-cloudsc-naml/build/venv_pyiface/lib/python3.11/site-packages/click/core.py", line 1130, in __call__
5:     return self.main(*args, **kwargs)
5:            ^^^^^^^^^^^^^^^^^^^^^^^^^^
5:   File "/Users/piotrows/dwarf-p-cloudsc-naml/build/venv_pyiface/lib/python3.11/site-packages/click/core.py", line 1055, in main
5:     rv = self.invoke(ctx)
5:          ^^^^^^^^^^^^^^^^
5:   File "/Users/piotrows/dwarf-p-cloudsc-naml/build/venv_pyiface/lib/python3.11/site-packages/click/core.py", line 1404, in invoke
5:     return ctx.invoke(self.callback, **ctx.params)
5:            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
5:   File "/Users/piotrows/dwarf-p-cloudsc-naml/build/venv_pyiface/lib/python3.11/site-packages/click/core.py", line 760, in invoke
5:     return __callback(*args, **kwargs)
5:            ^^^^^^^^^^^^^^^^^^^^^^^^^^^
5:   File "/Users/piotrows/dwarf-p-cloudsc-naml/build/bin/cloudsc_pyiface.py", line 67, in main
5:     path=input_path/'reference.h5', clsc=clsc, **npars
5:          ~~~~~~~~~~^~~~~~~~~~~~~~~
5: TypeError: unsupported operand type(s) for /: 'str' and 'str'
5/5 Test #5: dwarf-cloudsc-lib ............................***Failed    0.18 sec

File: src/cloudsc_pyiface/pyproject.toml
Comment (piotrows) [inline src/cloudsc_pyiface/pyproject.toml:1]: Should we credit you @mlange here ?

**Review by piotrows — CHANGES_REQUESTED** [review]
Thanks. The ctest still doesn't work for pyiface on my Mac, perhaps you can give a second look and see if my error messages match yours.

**Comment by piotrows** [comment]
@mlange @reuterbal  I have checked that the code works properly on Mac and ATOS login and interactive, both by ctest and direct execution. There is one serious and one minor problem I am seeing. The minor one is that Mac compilation, the following comment? is given while in CMake:
```-- Create Python virtual environment /Users/piotrows/dwarf-p-cloudsc-naml/build/venv_pyiface
sed: 1: "/Users/piotrows/dwarf-p ...": extra characters at the end of p command
-- Found Python3: /Users/piotrows/dwarf-p-cloudsc-naml/build/venv_pyiface/bin/python3.11 (found version "3.11.2") found components: Interpreter Development Development.Module Development.Embed```

The serious one is that the pyiface compilation does not work with nvhpc. Therefore, pyiface may need to be disabled by default. The error message given by f2py is:
```
nvc-Error-Unknown switch: -Wno-unused-result
nvc-Error-Unknown switch: -fwrapv
nvc-Error-nvc-Error-Unknown switch: -Wno-unused-resultUnknown switch: -Wno-unused-result
nvc-Error-nvc-Error-Unknown switch: -fwrapvUnknown switch: -fwrapv
```
i.e. compiler options appear in f2py from an unknown source.

reuterbal:
> ```
> sed: 1: "/Users/piotrows/dwarf-p ...": extra characters at the end of p command
> -- Found Python3: /Users/piotrows/dwarf-p-cloudsc-naml/build/venv_pyiface/bin/python3.11 (found version "3.11.2") found components: Interpreter Development Development.Module Development.Embed```
> ```

I have spotted this as well recently. For this use case, I recommend ignoring it for now (it will only make the portable shebang trick break). I'll likely replace the `sed` command by an inline Python regex replacement, since we have already a dependency on Python and that will make the solution portable.

PR #52, opened 2023-05-13
Python-gt4py: Removing GT4Py-based implementation of CLOUDSC

mlange05:
Due to licensing issues we need to remove this. Hopefully this can be re-introduced soon.


reuterbal (approved):
Thanks, looks good. Hoping this can be resolved soon.

PR #55, opened 2023-06-12
Enable gvmode and add nvidia/22.11 arch files

reuterbal:
This adds the gvmode (gang-vector-mode) flag for NVHPC on HPC2020, and adds updated arch files for nvidia/22.11

### DIFF
--- a/arch/ecmwf/hpc2020/nvhpc/22.11/env.sh
+++ b/arch/ecmwf/hpc2020/nvhpc/22.11/env.sh
@@ -1,0 +1,52 @@
+# (C) Copyright 1988- ECMWF.
+#
+# This software is licensed under the terms of the Apache Licence Version 2.0
+# which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
+# In applying this licence, ECMWF does not waive the privileges and immunities
+# granted to it by virtue of its status as an intergovernmental organisation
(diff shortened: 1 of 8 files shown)


mlange05 (changes requested):
Ok, in principle this looks great, but the `--with-cuda` mode breaks for me with the new env. I get:
```
/usr/bin/ld: cannot find -lcublas
/usr/bin/ld: cannot find -lcublasLt
/usr/bin/ld: cannot find -lcurand
```
in several places, which needs fixing before we can land this. I've verified the effect of `gvmode` though, and everything else looks super-neat and fast!

reuterbal:
Thanks for noticing this. Upon investigation, I could reproduce the issue and suspect a change in behaviour with respect to the `-cuda`/`-Mcuda` flag in NVHPC, possibly combined with specific target architecture specifications or CMake changes (which may have added this in the past automatically).

The specific issue, as far as I can tell, was that `cloudsc-common-lib` contains files with that option at compile time, for which nvfortran requires the option to be present at link time as well. However, the CMakeLists.txt specified this only for the build interface and was linked itself without it. Please verify if the latest commit resolves the issue for you.

mlange05:
Great, yes, all good now! Thanks for fixing this, GTG. :shipit:

mlange05 (approved):

PR #57, opened 2023-07-18
Bundle integration for Atlas

reuterbal:
As mentioned in the review of #54, this adds Atlas and its dependencies as optional projects to the bundle.
By default, they are not being built, which should allow to pick up an existing Atlas installation as usual. `--with-atlas` will enable building `eckit`, `fckit` and `atlas`. By default, tests are disabled for these dependencies but can be enabled using `--with-dependency-tests`.

I have also enabled Github actions integration for the new variant.

### DIFF
--- a/.github/scripts/verify-targets.sh
+++ b/.github/scripts/verify-targets.sh
@@ -17,7 +17,7 @@ fi
 
 if [[ "$gpu_flag" == "--with-gpu" ]]
 then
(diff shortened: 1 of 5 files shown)


wdeconinck (approved):

wdeconinck:
Merge pending ctest PR #56 
Then @sbrdar should rebase PR #54 
Then this PR should be rebased.

PR #66, opened 2023-12-13
SCC and SCC-K-CACHING STD PAR GPU offload

MichaelSt98:
STD parallelism/ `DO CONCURRENT` GPU offload (for NVIDIA GPUs)

* validates only for nvhpc 22.1, compiles and runs for nvhpc 22.11 as well but validation fails (for some reason)
* unable to use `cloudsc_add_compile_option()` or `target_compile_options()` to make it work (instead of changing the arch files)


CMakeLists.txt
```diff
@@ -54,6 +54,11 @@ ecbuild_add_option( FEATURE ACC
                     DESCRIPTION "Support for using GPUs with OpenACC"
                     REQUIRED_PACKAGES "OpenACC COMPONENTS Fortran" )
 
+ecbuild_add_option( FEATURE STDPAR
+                    DEFAULT ON
+                    DESCRIPTION "Support for using GPUs with OpenACC"
```
reuterbal:
```suggestion
                    DESCRIPTION "Support for using GPUs via DO CONCURRENT and OpenACC for global variable offloading"
```

.github/workflows/build-hpc.yml
```diff
@@ -55,6 +55,7 @@ jobs:
             modules: ~
             build_options: --with-loki=ON --with-field-api=ON
             # TODO: These tests trigger a SIGFPE, to be investigated
+            #  'dwarf-cloudsc-gpu-std-par-scc-k-caching-dp-serial' works with NVHPC 22.11 but not with 24.5
```
reuterbal: Is the failure with 24.5 also a SIGFPE or something different? Would be good to be explicit here.

reuterbal (changes requested):
Very nice, thanks for the updates and sorry for letting it hang for so long! Just two small changes I've noticed, and I'll take it for a spin on AC before merging.

reuterbal (approved):
Thanks, looks great. Can confirm it builds and runs on AC with NVHPC. I'm consistently seeing SIGFPE in single precision with both 22.11 and 24.5. Performance is rather bad, though, and I'm not getting any output with `NVCOMPILER_ACC_NOTIFY=1`, so I'm not sure if GPU offloading is actually happening?

Also, I noticed that I'm getting the following message:
```
free: cuMemFree returns error code 1
free: cuMemFree returns error code 1
free: cuMemFree returns error code 1
free: cuMemFree returns error code 1
```

PR #69, opened 2024-01-31
Loki F2C using `loki_transform` instead of `loki_transform_transpile`

MichaelSt98:
Necessary changes to make [Loki PR 208](https://github.com/ecmwf-ifs/loki/pull/208) work.


reuterbal (approved):
Thanks, looks good. As mentioned in ecmwf-ifs/loki#208, I suggest to merge the Loki PR first and see tests turn green here before merging.

PR #70, opened 2024-02-05
Update to new open-source FIELD_API

awnawab:
This PR updates the scc-field variant to use the new open-source FIELD_API. Notable changes are the use of the `FieldGang` feature to offload packed fields as a single contiguous entity, and the use of raw device pointers rather than host-mapped device pointers.


bundle.yml
```diff
@@ -69,14 +69,13 @@ projects :
...
-        version : master
+        git     : https://github.com/ecmwf-ifs/field_api.git
+        version : main
         optional: true
         require : ecbuild
         cmake   : >
-            ENABLE_FIELD_API_TESTS=OFF
-            ENABLE_FIELD_API_FIAT_BUILD=OFF
-            FIELD_API_UTIL_MODULE_PATH=${CMAKE_SOURCE_DIR}/cloudsc-dwarf/src/common/module
+            ENABLE_SINGLE_PRECISION=OFF
```
reuterbal, line 77: Since this is effectively a global option in this form, we might want to simply put this at the top of the bundle.yml.

reuterbal (changes requested):
Thanks, this looks great and very much looking forward to switching to the open-source field-api repository! Currently, the NVHPC tests fail to compile field-api and it would be good to get this sorted.

awnawab:
Converting this to a draft since I will need a FIELD_API build-system update to get around building with `acc=host` and `cuda` (for the CLOUSC CI).

awnawab:
Thanks again @reuterbal for all the help with this. The CI now builds the nvhpc runs with the `-acc=gpu` flag i.e. how we intend to build them for use. Because this inserts cuInit calls, we can't run these on the CI yet. Therefore, a separate nvhpc run is added with OpenACC explicitly disabled to test the loki and gpu variants (similar to the GNU CI runs).

reuterbal:
As discussed: Merging now and let's have a 1-line change PR once the new FIELD_API version has been tagged!

PR #62, opened 2023-12-05
Emancipation from serialbox via HDF5 for C-style variants

MichaelSt98:
**Draft/POC implementation HDF5 input and reference reading for C-style variants to not rely on serialbox for those variants.**

Before I am tempted to over-engineer this I thought I'd make a Draft PR to allow for some comments/feedback and to allow a discussion about a potential restructuring/refactoring of the C-style variants. 

Tested on 

* **AC** with Intel and NVHPC
* **LUMI** via cray toolchain 

Obviously more testing necessary before thinking about merging this.

Some hacks in the CMakeLists.txt due to problems with HDF5 on AC.

PR #71, opened 2024-02-13
HDF5 support for CUDA/HIP/SYCL variants

MichaelSt98:
HDF5 support for CUDA/HIP/SYCL variants, similar to PR #62 ...

CUDA and SYCL tested on AC with NVHPC 22.11

HIP tested on LUMI ... **which gave me some problems, explaining why this is a draft PR ...**

* unable to find `hdf5.h`
* able to find `hdf5.h` removing `set(CMAKE_CXX_COMPILER "${ROCM_PATH}/bin/hipcc")`
    * removing this allows still compilation for the device (at least for CCE 16.01)
    * however, different problems with CCE 15.01
  
**tested a lot of different things, without any success in finding a solution that works for the different toolchains/cray-gpu versions**


reuterbal (approved):
Confirmed working also on AC. Many thanks!

PR #72, opened 2024-02-20
Fix LUMI-G archs for HDF5 in HIP variants

reuterbal:
As discussed offline: env-file hack from IFS to introspect HDF5 module


MichaelSt98 (approved):
Thank you! I can confirm this works!

PR #73, opened 2024-03-19
Add semantic for reading thread count from $OMP_NUM_THREADS

antoine-morvan:
Hello, 

I am using some scripts to automatically launch the CloudSC dwarf with various numbers of OpenMP thread. The framework however relies on `$OMP_NUM_THREAD` to set this value, and cannot act on the executable arguments.

This PR does 2 things:
1. Prevent the application from crashing when the first argument is 0 or negative, by adding the semantic below. Previously it was failing on trying to start regions with 0 threads, or allocating arrays of size -1 (leading to some funny stuff with some compilers and -Ofast ...);
2. Add semantic to the value of the first argument when zero or negative: it now reads the `$OMP_NUM_THREADS` variable from the environment if present or defaults to CPU count (`omp_get_max_threads()`);

The updated readme should reflect these changes.

This should not break existing behavior, as zero or negative arguments were leading to aplication crash.

Best regards.


FussyDuck:
[CLA assistant check](https://bol-claassistant.ecmwf.int:443/ecmwf-ifs/dwarf-p-cloudsc?pullRequest=73) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://bol-claassistant.ecmwf.int:443/ecmwf-ifs/dwarf-p-cloudsc?pullRequest=73) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://bol-claassistant.ecmwf.int:443/check/ecmwf-ifs/dwarf-p-cloudsc?pullRequest=73) it.</sub>

src/cloudsc_c/dwarf_cloudsc.c
```diff
@@ -33,6 +33,9 @@ int main( int argc, char *argv[] ) {
     omp_threads = atoi( argv[1] );
     ngptot      = atoi( argv[2] );
     nproma      = atoi( argv[3] );
+    if (omp_threads <= 0) {
+      omp_threads = omp_get_max_threads();
+    }
```
reuterbal, line 38: Can we guard this with an `#ifdef _OPENMP` to retain the ability of building without OpenMP?

src/cloudsc_cuda/dwarf_cloudsc.cpp
```diff
@@ -33,6 +33,9 @@ int main( int argc, char *argv[] ) {
     omp_threads = atoi( argv[1] );
     ngptot      = atoi( argv[2] );
     nproma      = atoi( argv[3] );
+    if (omp_threads <= 0) {
+      omp_threads = omp_get_max_threads();
+    }
```
reuterbal, line 38: Also, guard with `#ifdef _OPENMP`?

src/cloudsc_fortran/dwarf_cloudsc.F90
```diff
@@ -19,6 +19,8 @@ PROGRAM DWARF_CLOUDSC
 USE YOMCST   , ONLY : YRCST 
 USE YOETHF   , ONLY : YRTHF
 
+USE OMP_LIB
+
```
reuterbal, line 23: Also, guard with `#ifdef _OPENMP`, and in l.53 below

src/cloudsc_gpu/dwarf_cloudsc_gpu.F90
```diff
@@ -42,6 +42,8 @@ PROGRAM DWARF_CLOUDSC
 USE CLOUDSC_DRIVER_GPU_SCC_FIELD_MOD, ONLY: CLOUDSC_DRIVER_GPU_SCC_FIELD
 #endif
 
+USE OMP_LIB
```
reuterbal, line 45: Guard with `#ifdef _OPENMP`, and in l.75 below

src/cloudsc_loki/dwarf_cloudsc.F90
```diff
@@ -18,6 +18,7 @@ PROGRAM DWARF_CLOUDSC
 USE CLOUDSC_DRIVER_LOKI_MOD, ONLY: CLOUDSC_DRIVER
 #endif
 USE EC_PMON_MOD, ONLY: EC_PMON
+USE OMP_LIB
```
reuterbal, line 21: Also, guard with `#ifdef _OPENMP`, and in l.51 below

reuterbal (changes requested):
Thanks, this looks like a good idea!

NB: I have retargeted the PR to develop

I haven't tested this myself, yet, but leaving a request for guarding the use of the OpenMP API, to make sure this can build without OpenMP available.

mlange05 (commented):
Hi @antoine-morvan many thanks for the contribution. 

I also agree that this is a good idea, but that we might also need the `#ifdef` guards. However, I've run this and recreated the previous failure, and it all works nicely and as expected, so should be good to go once this is in place.

antoine-morvan:
> Hi @antoine-morvan many thanks for the contribution.
> 
> I also agree that this is a good idea, but that we might also need the `#ifdef` guards. However, I've run this and recreated the previous failure, and it all works nicely and as expected, so should be good to go once this is in place.

Regarding the `#ifdef` in the C and CUDA implementations, the `cloudsc_drivers.h` files already includes `#include <omp.h>` without guard. The guard can be added nonetheless.

For the in Fortran files that would match what's in the `timer_mod`.  I'll add this.

src/cloudsc_c/dwarf_cloudsc.c
antoine-morvan: resolved in next commit

src/cloudsc_cuda/dwarf_cloudsc.cpp
antoine-morvan: resolved in next commit

src/cloudsc_fortran/dwarf_cloudsc.F90
antoine-morvan: resolved in next commit

src/cloudsc_gpu/dwarf_cloudsc_gpu.F90
antoine-morvan: resolved in next commit

src/cloudsc_loki/dwarf_cloudsc.F90
antoine-morvan: resolved in next commit

antoine-morvan:
Should be better with OpenMP-less support :)

mlange05 (approved):
Ok, fantastic. Many thanks @antoine-morvan  GTG from me.

reuterbal (approved):
Many thanks for adding the guards! We should sort this for CUDA/C as well, but that's for a different PR.

PR #74, opened 2024-03-21
Compatibility with Loki v0.2

reuterbal:
This will make CLOUDSC processing compatible with the new Scheduler in Loki v0.2. Once ecmwf-ifs/loki#213 is merged, we need this change to make it compatible with the new Loki version.


mlange05:
I filed a small fix that also adjusts the CUDA-specific loki config file - this is not tested but trips on AC NVHPC builds. Once that is in, this is GTG for me.

reuterbal:
Tests are green after merging the Loki PR.

mlange05 (approved):
All good for me. :shipit:

PR #75, opened 2024-03-21
Adjust CUDA-specific loki config file to Loki v0.2.0

mlange05:
This needs a few small special-case entries in the `disable` list.


reuterbal (approved):
Many thanks, I overlooked this!

PR #77, opened 2024-03-22
Update loki version to 0.2.0 and field_api to 0.3.0

reuterbal:
This is to fix the versions in the bundle in preparation for a new release.


mlange05 (approved):
Looks good.

PR #79, opened 2024-03-25
Loki: Add yoecldp.F90 header that contains NCLV to fix SCC demotion

mlange05:
If Loki doesn't recognise NCLV as a parameter it will not demote the solver matrix arrays correctly.

@reuterbal This might be related to the last minute Scheduler changes, or might have been broken for a little while (not sure). However, with this in place, we get and the other Loki change we get equivalent performance between the respective "loki" and "gpu" variants again.


mlange05:
Closing now, as superseded by #80.

PR #82, opened 2024-03-28
Ensure yoecldp.F90 is parsed with FParser

reuterbal:
An annoying oversight: OMNI has become broken with v1.5.1 due to yoecldp.F90 being now parsed (it was on the block list previously) but the OMNI frontend segfaulting on it... Testing did not pick up on this, because CLOUDSC tests use only Fparser, and Loki regression wasn't tested against the configuration changes until after the release.

This PR ensures that file is always parsed with FParser


mlange05 (approved):
LGTM, uncontroversial and GTG. :shipit:

PR #83, opened 2024-03-28
Mark projects no longer as optional and fix field-api to v0.3.0

reuterbal:
Another annoying oversight: The correct field-api tag name is `v0.3.0` (and not `0.3.0`). But the field-api project was marked as optional (to gracefully ignore that the field-api dependency could not be downloaded when it was still hosted internally). With field-api now available on Github, it should no longer be necessary to mark it as optional, ensuring we fail in the bundle creation when a tag is invalid.

### DIFF
--- a/bundle.yml
+++ b/bundle.yml
@@ -23,7 +23,6 @@ projects :
     - serialbox :
         git     : https://github.com/reuterbal/serialbox
         version : v2.5.4/patched
-        optional: true
         cmake   : >
             SERIALBOX_BUILD_SHARED=ON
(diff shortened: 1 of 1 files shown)


mlange05 (approved):
Ok, as discussed offline, I'm not a huge fan of forcing every repo to be checked out always, as some of them are large and only used for specific sub-options. However, ecbundle will always attempt to do so, only that it will fail silently without these flags, so this is beyond our control at this point. 

But regardless of ecbundle's limitations, this looks good as is for our testing pipeline. :shipit:

PR #84, opened 2024-04-11
Version 1.5.2

reuterbal:
This brings in the two technical changes overlooked in v1.5.1:
- Loki config file fix for use with the OMNI frontend
- Correct field_api tag v0.3.0 (instead of 0.3.0)


mlange05 (approved):
Looks great! Many thanks, and sorry for delay. GTG :shipit:

PR #85, opened 2024-04-26
Arch and CMake fixes from DAWN

reuterbal:
This amends the arch files for the two current compiler releases on DAWN and makes sure the relevant environment variables are set in the env.sh files.

Additionally, I have cleaned up the CMake files slightly, and removed the `-sp` suffix from the single precision variants.


reuterbal:
This has been superseded by #92

PR #43, opened 2023-02-28
Release 1.4.0

reuterbal:
This is to merge develop into main to roll release 1.4.0. Updated `VERSION` and `AUTHORS.md`

PR #89, opened 2024-05-29
a CLOUDSC version using Atlas GPU structures

sbrdar:
This is a version of Atlas GPU data structures for CLOUDSC which runs SCC, SCC_HOIST, SCC_K_CACHING kernels:
* the code is isolated in src/cloudsc_gpu_atlas and can be build with "--with-atlas ON --with-cuda ON" (on Nvidia) or "--with-atlas ON --with-hip ON" (on AMD GPUs) during the "./cloudsc-bundle build" command. The built executables carry 'atlas' in their names
* add support for packing of nD-array variables into an (n+1)D-array using atlas::MultiField
* allocate of hoist variables is left to atlas::pluto resource manager which implements various offload strategies: unpinned, pinned, pinned_pool, zero-copy, managed


src/cloudsc_gpu_atlas/dwarf_cloudsc_atlas.F90
```diff
@@ -92,7 +92,7 @@ PROGRAM DWARF_CLOUDSC
   ENDDO
 
   ! Validate the output against serialized reference data
-  IF (NITER == 1) CALL GLOBAL_ATLAS_STATE%VALIDATE(FSET, FSPACE, NGPTOTG)
+  ! IF (NITER == 1) CALL GLOBAL_ATLAS_STATE%VALIDATE(FSET, FSPACE, NGPTOTG)
```
wdeconinck: This should not be disabled

arch/eurohpc/lumi/cray-gpu/17.0.1/env.sh
```diff
@@ -1,4 +1,4 @@
-# (C) Copyright 1988- ECMWF.
+) Copyright 1988- ECMWF.
```
wdeconinck: Revert this change

arch/eurohpc/lumi/cray-gpu/17.0.1/env.sh
```diff
@@ -8,12 +8,36 @@
...
-  module load $1
+  if [ "${2:-""}" == "ECBUNDLE_CONFIGURE_ONLY" ]; then
+    if [ -n "${ECBUNDLE_CONFIGURE:-""}" ]; then
+      module load $1
+    else
+      echo " WARNING: Module $1 not loaded (only during configuration)"
+    fi
+  else
+    module load $1
+  fi
```
wdeconinck: This change is not needed.

src/cloudsc_gpu_atlas/cloudsc_driver_scc_mod.F90
```diff
@@ -0,0 +1,558 @@
...
+      & TENDENCY_LOC_T, TENDENCY_LOC_A, TENDENCY_LOC_Q, &
+      & PRAINFRAC_TOPRFZ, TENDENCY_LOC_CLD )
+
+    TYPE(PERFORMANCE_TIMER), INTENT(INOUT) :: TIMER
+    INTEGER(KIND=JPIM), INTENT(IN) :: NGPTOT
+    INTEGER(KIND=JPIM), INTENT(IN) :: NGPBLKS
+    INTEGER(KIND=JPIM), INTENT(IN) :: NLEV
+    INTEGER(KIND=JPIM), INTENT(IN) :: NPROMA
+    REAL(KIND=JPRB), INTENT(IN)   :: PTSPHY       ! Physics timestep
+    REAL(KIND=JPRB), CONTIGUOUS  :: PLCRIT_AER(:,:,:)
```
wdeconinck: I suggest to add INTENT for this and following arguments

src/cloudsc_gpu_atlas/cloudsc_driver_scc_mod.F90
```diff
@@ -0,0 +1,558 @@
...
+
+    TRACE = ATLAS_TRACE("cloudsc_driver_mod.F90", __LINE__, "CLOUDSC_DRIVER","COMPUTE")
+
+    FIELD = FSET%FIELD("PCLV")
+    FSPACE = FIELD%FUNCTIONSPACE()
+    NPROMA = FIELD%SHAPE(1)
+    NLEV = FSPACE%LEVELS()
+    NGPTOT = FSPACE%SIZE()
+
+    NGPBLKS = (NGPTOT / NPROMA) + MIN(MOD(NGPTOT,NPROMA), 1)
```
wdeconinck:
```suggestion
    NGPBLKS = FIELD%SHAPE(4)
```

src/cloudsc_gpu_atlas/cloudsc_driver_scc_mod.F90
```diff
@@ -0,0 +1,558 @@
...
+    call device_allocator%deallocate(ZQSICE)
+    call device_allocator%deallocate(ZFOEEWMT)
+    call device_allocator%deallocate(ZFOEEW)
+    call device_allocator%deallocate(ZFOEELIQT)
+end if
+#endif
+  END SUBROUTINE CLOUDSC_KERNEL
+
+
+  SUBROUTINE CLOUDSC_DRIVER(FSET, NUMOMP, NGPTOTG, KFLDX, PTSPHY)
```
wdeconinck: NGPTOTG, KFLDX are unused variables (except perhaps for a print) and can be refactored out.

Also not sure NUMOMP makes sense in the GPU context. Perhaps this should be forcefully set to 1, @reuterbal ?

src/cloudsc_gpu_atlas/cloudsc_driver_scc_mod.F90
```diff
@@ -0,0 +1,558 @@
...
+    ! Local copy of cloud parameters for offload
+    TYPE(TECLDP) :: LOCAL_YRECLDP
+
+    HOIST_POOL = 0
+    CALL GET_ENVIRONMENT_VARIABLE("HOIST_POOL", VALUE=HOIST_POOL_STR, LENGTH=HOIST_POOL_STR_LEN)
+    IF (HOIST_POOL_STR_LEN > 0 ) THEN
+      READ(HOIST_POOL_STR(1:HOIST_POOL_STR_LEN), *) HOIST_POOL
+    ENDIF
+    IF (ALL(HOIST_POOL/=[-1,0,1])) THEN
+        PRINT *, "HOIST_POOL must be -1, 0, or 1. Exiting."
```
wdeconinck: I'd remove the '-1' option for HOIST_POOL, and don't do the `acc enter data create`, and only use pluto.

src/cloudsc_gpu_atlas/cloudsc_driver_scc_mod.F90
```diff
@@ -0,0 +1,558 @@
...
+
+#ifdef CLOUDSC_GPU_SCC_HOIST
+    ! Local declarations of promoted temporaries via atlas::pluto
+    TYPE(pluto_allocator) :: device_allocator
+    REAL(KIND=JPRB), POINTER :: ZFOEALFA(:,:,:), ZTP1(:,:,:), &
+        & ZLI(:,:,:), ZA(:,:,:), ZAORIG(:,:,:), ZLIQFRAC(:,:,:), &
+        & ZICEFRAC(:,:,:), ZQSMIX(:,:,:), ZQSLIQ(:,:,:), &
+        & ZQSICE(:,:,:), ZFOEEWMT(:,:,:), ZFOEEW(:,:,:), ZFOEELIQT(:,:,:)
+    REAL(KIND=JPRB), POINTER :: ZQX(:,:,:,:), ZQX0(:,:,:,:), &
+        & ZPFPLSX(:,:,:,:), ZLNEG(:,:,:,:), ZQXN2D(:,:,:,:)
```
wdeconinck, line 137: This can be factored out as a CLOUD_SCC_HOIST_DATA type, (including the TYPE(pluto_allocator))
Then the lines below to allocate and deallocate could be member functions of this type.

wdeconinck (changes requested):
Thanks @sbrdar for showing the power of atlas and pluto.
I have some suggestions.

src/cloudsc_fortran_atlas/cloudsc_global_atlas_state_mod.F90
```diff
@@ -266,6 +269,7 @@ SUBROUTINE CLOUDSC_GLOBAL_ATLAS_STATE_LOAD(SELF, FSET, FSPACE, NPROMA, NGPTOTG)
       MULTIFIELD_IN = atlas_MultiField(CONFIG)
       CALL FSET%ADD(MULTIFIELD_IN%FIELDSET()) 
       CALL MULTIFIELD_IN%FINAL()
+      PRINT *, "ERROR: MultiField not avaiable in Atlas/develop"
```
wdeconinck: Correction, it is now :)

src/cloudsc_gpu_atlas/dwarf_cloudsc_atlas.F90
```diff
@@ -0,0 +1,118 @@
...
+
+USE PARKIND1, ONLY: JPIM
+USE CLOUDSC_MPI_MOD, ONLY: CLOUDSC_MPI_INIT, CLOUDSC_MPI_END, NUMPROC, IRANK
+USE CLOUDSC_GLOBAL_ATLAS_STATE_MOD, ONLY: CLOUDSC_GLOBAL_ATLAS_STATE
+USE CLOUDSC_DRIVER_SCC_MOD, ONLY: CLOUDSC_DRIVER
+
+USE ATLAS_MODULE
+USE, INTRINSIC :: ISO_C_BINDING
+USE ATLAS_FUNCTIONSPACE_BLOCKSTRUCTUREDCOLUMNS_MODULE
+USE ATLAS_MULTIFIELD_MODULE
```
wdeconinck:
```suggestion
```
can be removed.

src/cloudsc_gpu_atlas/CMakeLists.txt
```diff
@@ -0,0 +1,125 @@
...
+            cloudsc_global_atlas_state_mod.F90
+            expand_atlas_mod.F90
+            validate_atlas_mod.F90
+            ../cloudsc_gpu/cloudsc_gpu_scc_hoist_mod.F90
+            cloudsc_driver_scc_mod.F90
+            dwarf_cloudsc_atlas.F90
+        LIBS
+            atlas_f
+            pluto_f
+            cloudsc-common-${prec}-atlas
```
wdeconinck:
```suggestion
            cloudsc-common-atlas-${prec}
```
I'd assume that ${prec} should go at the end of the name. Of course also change everywhere used and defined.

src/cloudsc_gpu_atlas/dwarf_cloudsc_atlas.F90
sbrdar: Enabled again to execute every time after the first iteration.

arch/eurohpc/lumi/cray-gpu/17.0.1/env.sh
sbrdar: Done. Thanks!

arch/eurohpc/lumi/cray-gpu/17.0.1/env.sh
sbrdar: Thanks! It has been removed.

src/cloudsc_gpu_atlas/cloudsc_driver_scc_mod.F90
sbrdar: Thanks. I have added the INTENTs.

src/cloudsc_gpu_atlas/cloudsc_driver_scc_mod.F90
sbrdar: Good point to avoid inconsistencies. Thanks!

src/cloudsc_gpu_atlas/cloudsc_driver_scc_mod.F90
sbrdar: I have removed both unused arguments in the CLOUDSC_DRIVER subroutine.

src/cloudsc_gpu_atlas/cloudsc_driver_scc_mod.F90
sbrdar: I have removed the '-1' option, we rely on 'pluto' to allocate the hoisting variables.

src/cloudsc_gpu_atlas/cloudsc_driver_scc_mod.F90
sbrdar: This approach is nice but it drifts away from the rest of the code with i.e. FIELD_API. Perhaps we can think of it together with @reuterbal?

src/cloudsc_fortran_atlas/cloudsc_global_atlas_state_mod.F90
sbrdar: Removed and also replaced the old usage of MultiField creation in the CPU version with the new one, similar to how it is done for the GPU version.

src/cloudsc_gpu_atlas/dwarf_cloudsc_atlas.F90
sbrdar: Thanks. It has been removed now.

src/cloudsc_gpu_atlas/CMakeLists.txt
sbrdar: This is the design of @reuterbal to have libraries, like atlas etc, attached to cloudsc-common after the precision. If @reuterbal agrees to change it in his build system, I have no problem changing it in the above CMake file.

src/cloudsc_gpu_atlas/cloudsc_driver_scc_mod.F90
wdeconinck: Sure you could leave it.

wdeconinck:
I would advise to reuse CPU and GPU codes where possible, to avoid code duplication and maintainability.
In particular, I think the validate, and creation of the fields and loading the data is probably identical.
The CPU OpenMP driver or GPU OpenACC driver can then be called independently.

sbrdar:
> I would advise to reuse CPU and GPU codes where possible, to avoid code duplication and maintainability. In particular, I think the validate, and creation of the fields and loading the data is probably identical. The CPU OpenMP driver or GPU OpenACC driver can then be called independently.

I have now reused the expand_atlas_mod and validate_atlas_mod from 'cloudsc_fortran_atlas' in 'cloudsc_gpu_atlas'.

arch/toolchains/ecmwf-hpc2020-nvhpc.cmake
```diff
@@ -30,6 +30,8 @@ set( OpenMP_C_LIB_NAMES     "acchost" CACHE STRING "")
 # NB: We have to add `-mp` again to avoid undefined symbols during linking
 # (smells like an Nvidia bug)
 set( OpenACC_Fortran_FLAGS "-acc=gpu -mp=gpu -gpu=cc80,lineinfo,fastmath,gvmode" CACHE STRING "" )
+set( OpenACC_C_FLAGS "${OpenACC_Fortran_FLAGS}")
```
reuterbal: should this not be a CACHE variable, too?

bundle.yml
```diff
@@ -104,6 +106,8 @@ options :
         cmake : >
             DWARF-P-CLOUDSC_ENABLE_ACC={{value}}
             FIELD_API_ENABLE_ACC={{value}}
+            ATLAS_ENABLE_ACC={{value}}
+            ATLAS_ENABLE_CUDA={{value}}
```
reuterbal: Is CUDA required, or does it work also with only OpenACC? Since the CUDA support can be enabled separately via `--with-cuda`, it would make more sense to restrict this to ATLAS_ENABLE_ACC here if possible.

src/cloudsc_gpu_atlas/CMakeLists.txt
```diff
@@ -0,0 +1,125 @@
...
+
+ecbuild_add_option( FEATURE CLOUDSC_GPU_SCC_HOIST_ATLAS
+    DESCRIPTION "Build optimized GPU version of CLOUDSC using SCC with hoisted temporary arrays (based on Atlas + OpenACC)" DEFAULT ON
+    CONDITION atlas_FOUND AND (Serialbox_FOUND OR HDF5_FOUND)
+)
+
+ecbuild_add_option( FEATURE CLOUDSC_GPU_SCC_K_CACHING_ATLAS
+    DESCRIPTION "Build optimized GPU version of CLOUDSC using SCC with k-caching of local array in the kernel (based on Atlas + OpenACC)" DEFAULT ON
+    CONDITION atlas_FOUND AND (Serialbox_FOUND OR HDF5_FOUND)
+)
```
reuterbal: We have retired the per-variant ecbuild options because it did get a little out of hand, making it hard to spot what options or packages were selected or not found. Instead, we build all viable variants that are possible with the found dependencies. Since the include in `src/CMakeLists.txt` already ensures that Atlas is available, we can skip the ecbuild_add_option here and the `if (HAVE_CLOUDSC_GPU_SCC_ATLAS)`, `if(HAVE_CLOUDSC_GPU_SCC_HOIST_ATLAS)` etc. below.

src/cloudsc_gpu_atlas/CMakeLists.txt
```diff
@@ -0,0 +1,125 @@
...
+
+if( HAVE_CLOUDSC_GPU_SCC_ATLAS )
+    ecbuild_add_executable(
+        TARGET dwarf-cloudsc-gpu-scc-atlas-${prec}
+        SOURCES
+            cloudsc_global_atlas_state_mod.F90
+            ../cloudsc_fortran_atlas/expand_atlas_mod.F90
+            ../cloudsc_fortran_atlas/validate_atlas_mod.F90
```
reuterbal: I think we should maybe move these into the common lib at some point? 

In any case, since they are built for every target here, you get problems with aliasing in the .mod files and potential race conditions in parallel builds. Please try a build with Ninja, which is great at catching these issues:
```
ninja: build stopped: multiple rules generate cloudsc-dwarf/src/cloudsc_gpu_atlas.dp/cloudsc_global_atlas_state_mod.mod.
```

Therefore, as a minimum you will need to set per-target module directories here. See for example the relevant snippet from the GPU variants:

https://github.com/ecmwf-ifs/dwarf-p-cloudsc/blob/b42aacae409f3a7bec13bdf0f4c75bee5f5989ba/src/cloudsc_gpu/CMakeLists.txt#L35-L38

src/CMakeLists.txt
```diff
@@ -48,6 +48,9 @@ foreach( prec IN LISTS CLOUDSC_PRECISIONS )
   endif()
   if( HAVE_ATLAS )
     add_subdirectory( cloudsc_fortran_atlas cloudsc_fortran_atlas.${prec} )
+    if( HAVE_CUDA OR HAVE_HIP )
```
reuterbal: I'm wondering, are CUDA/HIP really required to build these variants? It's not performant but a CPU build should work and be runnable, no?

src/cloudsc_gpu_atlas/CMakeLists.txt
reuterbal: Either is fine with me - I wasn't particularly strict about the naming scheme there, only trying to avoid any clashes and limiting the programming model exposure in the common libs.

src/cloudsc_gpu_atlas/CMakeLists.txt
```diff
@@ -0,0 +1,125 @@
...
+      ${CMAKE_CURRENT_SOURCE_DIR}/../../data ${CMAKE_CURRENT_BINARY_DIR}/../../../data )
+endif()
+
+
+if( HAVE_HDF5 )
+  execute_process(COMMAND ${CMAKE_COMMAND} -E create_symlink
+      ${CMAKE_CURRENT_SOURCE_DIR}/../../config-files/input.h5 ${CMAKE_CURRENT_BINARY_DIR}/../../../input.h5 )
+  execute_process(COMMAND ${CMAKE_COMMAND} -E create_symlink
+      ${CMAKE_CURRENT_SOURCE_DIR}/../../config-files/reference.h5 ${CMAKE_CURRENT_BINARY_DIR}/../../../reference.h5 )
+endif()
```
reuterbal: This is redundant, please remove. Test data is being handled centrally in CMake now.

src/cloudsc_gpu_atlas/CMakeLists.txt
```diff
@@ -0,0 +1,125 @@
...
+
+
+# Importantly, we add the "pinned" flag to the linker command to ensure pinning!
+#if( CMAKE_Fortran_COMPILER_ID MATCHES "NVHPC" )
+#    target_link_options( dwarf-cloudsc-gpu-scc-atlas PUBLIC "-gpu=pinned" )
+#    target_link_options( dwarf-cloudsc-gpu-scc-hoist-atlas PUBLIC "-gpu=pinned" )
+#    target_link_options( dwarf-cloudsc-gpu-scc-k-caching-atlas PUBLIC "-gpu=pinned" )
+#endif()
```
reuterbal: Can be deleted?

arch/ecmwf/hpc2020/nvhpc/22.11/env.sh
```diff
@@ -24,10 +24,12 @@ module_unload() {
 module purge
 
 # Load modules
+module_load cmake/3.25.2
```
reuterbal: CMake 3.28 is loaded further down
```suggestion
module_load cmake/3.25.2
```

src/cloudsc_gpu_atlas/CMakeLists.txt
```diff
@@ -0,0 +1,125 @@
...
+            pluto_f
+            cloudsc-common-${prec}-atlas
+        DEFINITIONS ${CLOUDSC_DEFINITIONS} CLOUDSC_GPU_SCC
+    )
+
+    ecbuild_add_test(
+        TARGET dwarf-cloudsc-gpu-scc-atlas
+        COMMAND bin/dwarf-cloudsc-gpu-scc-atlas
+        ARGS 1 1000 128
+        WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/../../..
```
reuterbal: This is where the central input file mechanics will place the test inputs. Please change this also for the other test definitions in this file.
```suggestion
        WORKING_DIRECTORY ${CMAKE_BINARY_DIR}
```

src/cloudsc_gpu_atlas/CMakeLists.txt
```diff
@@ -0,0 +1,125 @@
...
+        LIBS
+            atlas_f
+            pluto_f
+            cloudsc-common-${prec}-atlas
+        DEFINITIONS ${CLOUDSC_DEFINITIONS} CLOUDSC_GPU_SCC_HOIST
+    )
+
+    ecbuild_add_test(
+        TARGET dwarf-cloudsc-gpu-scc-hoist-atlas
+        COMMAND bin/dwarf-cloudsc-gpu-scc-hoist-atlas
```
reuterbal:
```suggestion
        TARGET dwarf-cloudsc-gpu-scc-hoist-atlas-${prec}
        COMMAND bin/dwarf-cloudsc-gpu-scc-hoist-atlas-${prec}
```

src/cloudsc_gpu_atlas/CMakeLists.txt
```diff
@@ -0,0 +1,125 @@
...
+        LIBS
+            atlas_f
+            pluto_f
+            cloudsc-common-${prec}-atlas
+            DEFINITIONS ${CLOUDSC_DEFINITIONS} CLOUDSC_GPU_SCC_K_CACHING
+    )
+
+    ecbuild_add_test(
+        TARGET dwarf-cloudsc-gpu-scc-k-caching-atlas
+        COMMAND bin/dwarf-cloudsc-gpu-scc-k-caching-atlas
```
reuterbal:
```suggestion
        TARGET dwarf-cloudsc-gpu-scc-k-caching-atlas-${prec}
        COMMAND bin/dwarf-cloudsc-gpu-scc-k-caching-atlas-${prec}
```

src/cloudsc_gpu_atlas/CMakeLists.txt
```diff
@@ -0,0 +1,125 @@
...
+        LIBS
+            atlas_f
+            pluto_f
+            cloudsc-common-${prec}-atlas
+        DEFINITIONS ${CLOUDSC_DEFINITIONS} CLOUDSC_GPU_SCC
+    )
+
+    ecbuild_add_test(
+        TARGET dwarf-cloudsc-gpu-scc-atlas
+        COMMAND bin/dwarf-cloudsc-gpu-scc-atlas
```
reuterbal: Tests don't pass because target name doesn't match:
```suggestion
        TARGET dwarf-cloudsc-gpu-scc-atlas-${prec}
        COMMAND bin/dwarf-cloudsc-gpu-scc-atlas-${prec}
```

src/cloudsc_gpu_atlas/cloudsc_driver_scc_mod.F90
```diff
@@ -0,0 +1,491 @@
...
+    INTEGER(KIND=JPIM) :: HOIST_POOL
+
+    HOIST_POOL = 0
+    CALL GET_ENVIRONMENT_VARIABLE("HOIST_POOL", VALUE=HOIST_POOL_STR, LENGTH=HOIST_POOL_STR_LEN)
```
reuterbal, line 140: Can you document this enviornment variable in the readme please?

arch/ecmwf/gh/env.sh
reuterbal, line 1: The env doesn't fully work on the Grace Hopper node, the prototype1 variant and the CUDA variants throw various compiler errors for me. Have you seen the same?

I'm ok with integrating it nonetheless since it been used to test the Atlas variant, but maybe add a print to the env.sh file that clearly states that the CUDA C and prototype1 variants are known to not work with this.

reuterbal (changes requested):
Many thanks for this, this is an exciting step forward and I'm very much looking forward to bringing this in! Apologies for the long review time, I did want to do this properly and really test this with the appropriate care.

First of all, the SCC and SCC-K-CACHING variants work flawlessly in single and double precision and also on larger problem sizes and produce very nice performance results! Well done!

- [ ] I left some remarks in-line, please address them. Most of the comments relate to the CMake integration, which needs some improvements. Also, I would like to make sure the tests work correctly and ideally switch Atlas on for them in a subsequent PR both in hpc-ci and at least on GH Actions configuration.

- [ ] For some reason, the per-thread throughput reporting (which should measure the kernel execution time) seems to not work correctly, e.g., on gh1 it simply shows 0:

```
$ bin/dwarf-cloudsc-gpu-scc-k-caching-atlas-dp 1 163840 128
Warning! Processor does not support trapping of floating-point exceptions.
 LMULTIFIELD =   F
 NGPTOTG     =        163840
 NGPTOT      =        163840
 NBLKS       =          1280
 NPROMA      =           128
 Host allocate [ms]:             9
 LMULTIFIELD =   F
 NGPTOTG     =        163840
 NGPTOT      =        163840
 NBLKS       =          1280
 NPROMA      =           128
 Host allocate [ms]:             1

### ITERATION 1/1
     NUMPROC=1, NUMOMP=1, NGTOT=163840, NGPTOTG=163840, NPROMA=128, NGPBLKS=1280
 Device allocate [ms]:            32
 Reference MFLOP count for 100 columns :  12.48232900
     NUMOMP    NGPTOT  #GP-cols     #BLKS    NPROMA tid# : Time(msec)  MFlops/s     col/s
          1    163840         0         0       128    0 :         21         0   7784519 @ rank#0:core#62
          1    163840         0      1280       128   -1 :         74    275665   2208443 : TOTAL @ rank#0
    1 x   1    163840         0      1280       128   -1 :         74    275665   2208443 : TOTAL
```

- [x] Finally, after fixing the issues mentioned in the comments, the following test doesn't work for me right now on AC GPUs (both in single and double precision and with and without OpenMP):

```
21/46 Test: dwarf-cloudsc-fortran-atlas-dp-serial
Command: "/.../nab_atlas_gpu_field/build/bin/dwarf-cloudsc-fortran-atlas-dp" "1" "100" "16"
Directory: /.../nab_atlas_gpu_field/build
"dwarf-cloudsc-fortran-atlas-dp-serial" start time: May 26 06:09 CEST
Output:
----------------------------------------------------------
  LMULTIFIELD:   T
(E) Exception: FieldSet: cannot find field "PLCRIT_AER"  (/.../nab_atlas_gpu_field/source/atlas/src/atlas/field/FieldSet.cc:120 field)

(E) =========================================
(E) TERMINATING dwarf-cloudsc-fortran-atlas-dp
(E) -----------------------------------------
(E) FieldSet: cannot find field "PLCRIT_AER"
(E) -----------------------------------------
(E) LOCATION:  (/.../nab_atlas_gpu_field/source/atlas/src/atlas/field/FieldSet.cc:120 field)
(E) -----------------------------------------
(E) BACKTRACE
(E) ------------------
```

And the SCC-HOIST test fails with a segfault:
```
24/46 Test: dwarf-cloudsc-gpu-scc-hoist-atlas-dp
Command: "/.../nab_atlas_gpu_field/build/bin/dwarf-cloudsc-gpu-scc-hoist-atlas-dp" "1" "1000" "128"
Directory: /.../nab_atlas_gpu_field/build/cloudsc-dwarf/src/cloudsc_gpu_atlas.dp/../../..
"dwarf-cloudsc-gpu-scc-hoist-atlas-dp" start time: May 26 06:09 CEST
Output:
----------------------------------------------------------
 LMULTIFIELD =   F
 NGPTOTG     =          1000
 NGPTOT      =          1000
 NBLKS       =             8
 NPROMA      =           128
 LMULTIFIELD =   F
 NGPTOTG     =          1000
 NGPTOT      =          1000
 NBLKS       =             8
 NPROMA      =           128

### ITERATION 1/1
     NUMPROC=1, NUMOMP=1, NGTOT=1000, NGPTOTG=1000, NPROMA=128, NGPBLKS=8
<end of output>
Test time =   0.38 sec
----------------------------------------------------------
Test Failed.
```

arch/toolchains/ecmwf-hpc2020-nvhpc.cmake
sbrdar: You are right, I did not find OpenACC_C_FLAGS in the cache list and it should be there. I am adding CACHE here.

bundle.yml
sbrdar: Good point. CUDA is not required on --with-acc, Atlas can resolve now itself what it needs in the background.

src/cloudsc_gpu_atlas/CMakeLists.txt
sbrdar: Indeed, cmake was more involved with condition testing. I am happy to clean up all add_option-s now, assuming the conditions are satisfied.

src/cloudsc_gpu_atlas/CMakeLists.txt
sbrdar: A valuable comment, much appreciated. I have at the moment opted for the minimum after your suggestion, which seems to work.

src/CMakeLists.txt
sbrdar: It is true. We can compile and run Atlas GPU code for CPU. I have removed the CUDA/HIP requirement here, albeit without thoroughly testing.

src/cloudsc_gpu_atlas/CMakeLists.txt
sbrdar: Thanks. Cleaned up now.

src/cloudsc_gpu_atlas/CMakeLists.txt
sbrdar: Thanks. Removed now.

arch/ecmwf/hpc2020/nvhpc/22.11/env.sh
sbrdar: Indeed. I have removed 3.25.2 from the list.

src/cloudsc_gpu_atlas/CMakeLists.txt
sbrdar: Done. Thanks.

src/cloudsc_gpu_atlas/CMakeLists.txt
sbrdar: Cleaned up now. Thanks!

src/cloudsc_gpu_atlas/CMakeLists.txt
sbrdar: ... and this one as well.

src/cloudsc_gpu_atlas/CMakeLists.txt
sbrdar: ... and this one as well.

src/cloudsc_gpu_atlas/cloudsc_driver_scc_mod.F90
sbrdar: Thanks. Done. Also, added information on dwarf-cloudsc-gpu-scc-k-caching-atlas.

src/cloudsc_gpu_atlas/CMakeLists.txt
```diff
@@ -7,119 +7,91 @@
...
+            Fortran_MODULE_DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}/module-dwarf-cloudsc-gpu-scc-hoist-atlas"
+)
 
+ecbuild_add_test(
+    TARGET dwarf-cloudsc-gpu-scc-hoist-atlas-${prec}
+    COMMAND bin/dwarf-cloudsc-gpu-scc-hoist-atlas-${prec}
+    ARGS 1 1000 128
+    WORKING_DIRECTORY ${CMAKE_BINARY_DIR}
+    OMP 1
+)
```
reuterbal: Sorry for not spotting this earlier. Can we add a second test that enables the hoist pool?
```suggestion
ecbuild_add_test(
    TARGET dwarf-cloudsc-gpu-scc-hoist-atlas-${prec}
    COMMAND bin/dwarf-cloudsc-gpu-scc-hoist-atlas-${prec}
    ARGS 1 1000 128
    WORKING_DIRECTORY ${CMAKE_BINARY_DIR}
    OMP 1
)
ecbuild_add_test(
    TARGET dwarf-cloudsc-gpu-scc-hoist-pool-atlas-${prec}
    COMMAND bin/dwarf-cloudsc-gpu-scc-hoist-atlas-${prec}
    ARGS 1 1000 128
    WORKING_DIRECTORY ${CMAKE_BINARY_DIR}
    OMP 1
    ENVIRONMENT HOIST_POOL=1
)
```

src/cloudsc_fortran_atlas/cloudsc_global_atlas_state_mod.F90
```diff
@@ -209,23 +212,21 @@ SUBROUTINE CLOUDSC_GLOBAL_ATLAS_STATE_LOAD(SELF, FSET, FSPACE, NPROMA, NGPTOTG)
...
     INTEGER :: CENV_LEN
     TYPE(ATLAS_TRACE) :: TRACE, TRACE_IO
+    LOGICAL :: LMULTIFIELD = .FALSE.
 
     TRACE = ATLAS_TRACE("cloudsc_global_atlas_state_mod.F90", __LINE__, &
         & "CLOUDSC_GLOBAL_ATLAS_STATE_LOAD")
 
-    LMULTIFIELD = .TRUE.
-    CALL GET_ENVIRONMENT_VARIABLE("CLOUDSC_ATLAS_MULTIFIELD",CENV,CENV_LEN)
+    CALL GET_ENVIRONMENT_VARIABLE("CLOUDSC_ATLAS_MULTIFIELD", CENV, CENV_LEN)
```
reuterbal, line 224: Would you mind also adding a second test with `CLOUDSC_ATLAS_MULTIFIELD=1`?

reuterbal (changes requested):
Many thanks for addressing all the comments @sbrdar! This looks very good now.

I just left two small requests inline to add test variants for the options introduced via environment variables to make sure these features are tested appropriately.

One thing I noticed, however, is that when not specifying the `--with-cuda=ON` option, the atlas-scc and atlas-scc-hoist variants seem to segfault. But tests work fine with the `--with-cuda=ON` option. Tested with nvidia/24.5 on AC:

```sh
srun -c 64 ./cloudsc-bundle build --arch arch/ecmwf/hpc2020/nvhpc/24.5/ --with-atlas=ON --build-dir build.nvidia [--with-cuda=ON]
cd build.nvidia
source env.sh
srun -q dg -t 0:30:00 -n 1 -c 32 --mem=120G --gpus=1 ctest
```

Can you please have another look what might be the issue here?

sbrdar:
> Many thanks for addressing all the comments @sbrdar! This looks very good now.
> 
> I just left two small requests inline to add test variants for the options introduced via environment variables to make sure these features are tested appropriately.
> 
> One thing I noticed, however, is that when not specifying the `--with-cuda=ON` option, the atlas-scc and atlas-scc-hoist variants seem to segfault. But tests work fine with the `--with-cuda=ON` option. Tested with nvidia/24.5 on AC:
> 
> ```shell
> srun -c 64 ./cloudsc-bundle build --arch arch/ecmwf/hpc2020/nvhpc/24.5/ --with-atlas=ON --build-dir build.nvidia [--with-cuda=ON]
> cd build.nvidia
> source env.sh
> srun -q dg -t 0:30:00 -n 1 -c 32 --mem=120G --gpus=1 ctest
> ```
> 
> Can you please have another look what might be the issue here?

We have not considered this so far, but it has been fixed now. A partial FE_INVALID check on the CLOUDSC function had to be omitted in the release build when running on CPU. Also, dwarf files using Atlas data should not use OpenACC if Atlas does not have CUDA/HIP enabled. OpenACC accesses Atlas data structured through the 'device_ptr' pragma call.

src/cloudsc_gpu_atlas/CMakeLists.txt
sbrdar: Thanks for the hint. It has been added.

src/cloudsc_fortran_atlas/cloudsc_global_atlas_state_mod.F90
sbrdar: ... and this as well. Thank you!

src/cloudsc_gpu_atlas/cloudsc_driver_scc_mod.F90
```diff
@@ -182,6 +183,10 @@ SUBROUTINE CLOUDSC_KERNEL( TIMER, NGPTOT, NGPBLKS, NLEV, NPROMA, PTSPHY, &
   call device_allocator%allocate(ZFOEELIQT, [NPROMA, NLEV, NGPBLKS])
 #endif
 
+!#if (ATLAS_HAVE_GPU == 0)
+    CALL IEEE_SET_HALTING_MODE(IEEE_INVALID, .FALSE.)
+!#endif
```
wdeconinck:
```suggestion
    LOGICAL :: LL_HALT_INVALID
    CALL IEEE_GET_HALTING_MODE(IEEE_INVALID, LL_HALT_INVALID)
    CALL IEEE_SET_HALTING_MODE(IEEE_INVALID, .FALSE.)
```
Also please write a little comment stating that this temporarily disables floating-point-trapping of FE_INVALID caused by too aggressive optimisation within the CLOUDSC function.

src/cloudsc_gpu_atlas/cloudsc_driver_scc_mod.F90
```diff
@@ -253,6 +258,10 @@ SUBROUTINE CLOUDSC_KERNEL( TIMER, NGPTOT, NGPBLKS, NLEV, NPROMA, PTSPHY, &
 !$acc end parallel loop
 !$acc end data
 
+!#if (ATLAS_HAVE_GPU == 0)
+    CALL IEEE_SET_HALTING_MODE(IEEE_INVALID, .TRUE.)
+!#endif
```
wdeconinck:
```suggestion
    IF (LL_HALT_INVALID) CALL IEEE_SET_HALTING_MODE(IEEE_INVALID,  .TRUE.)
```
Only set it back to `.TRUE.` if it was `.TRUE.` before (`LL_HALT_INVALID`)

bundle.yml
```diff
@@ -73,6 +73,8 @@ projects :
         cmake   : >
             ATLAS_ENABLE_TESTS=OFF
             ATLAS_ENABLE_CUDA=OFF
+            ATLAS_ENABLE_HIP=OFF
+            ATLAS_ENABLE_ACC=OFF
```
wdeconinck: CUDA, HIP, ACC are OFF by default, so no need to add here.

src/common/CMakeLists.txt
```diff
@@ -131,14 +131,23 @@ endif()
...
 
 if( HAVE_ATLAS )
+    set( ATLAS_HAVE_GPU 0 )
+    if ( pluto_HAVE_CUDA OR pluto_HAVE_HIP )
+        set( ATLAS_HAVE_GPU 1 )
+    endif()
+    set( USE_ACC ${HAVE_ACC} )
+    if ( NOT ATLAS_HAVE_GPU )
+       set( USE_ACC 0 )
+    endif()
```
wdeconinck, line 142: Write a comment that Atlas was not compiled with GPU support, so that OpenACC instructions should be ignored in the dwarf

wdeconinck (changes requested):

src/common/CMakeLists.txt
sbrdar: Good point. I have added a cmake warning on disabling OpenACC instruction when GPU support has been disabled in Atlas.

bundle.yml
sbrdar: Indeed, they are off in Atlas. Hence, removed here. Thanks!

src/cloudsc_gpu_atlas/cloudsc_driver_scc_mod.F90
sbrdar: Fixed. Thanks!

src/cloudsc_gpu_atlas/cloudsc_driver_scc_mod.F90
sbrdar: Many thanks. It has been added.

sbrdar:
> Many thanks for this, this is an exciting step forward and I'm very much looking forward to bringing this in! Apologies for the long review time, I did want to do this properly and really test this with the appropriate care.
> 
> First of all, the SCC and SCC-K-CACHING variants work flawlessly in single and double precision and also on larger problem sizes and produce very nice performance results! Well done!
> 
> * [ ]  I left some remarks in-line, please address them. Most of the comments relate to the CMake integration, which needs some improvements. Also, I would like to make sure the tests work correctly and ideally switch Atlas on for them in a subsequent PR both in hpc-ci and at least on GH Actions configuration.
> * [ ]  For some reason, the per-thread throughput reporting (which should measure the kernel execution time) seems to not work correctly, e.g., on gh1 it simply shows 0:
> 
> ```
> $ bin/dwarf-cloudsc-gpu-scc-k-caching-atlas-dp 1 163840 128
> Warning! Processor does not support trapping of floating-point exceptions.
>  LMULTIFIELD =   F
>  NGPTOTG     =        163840
>  NGPTOT      =        163840
>  NBLKS       =          1280
>  NPROMA      =           128
>  Host allocate [ms]:             9
>  LMULTIFIELD =   F
>  NGPTOTG     =        163840
>  NGPTOT      =        163840
>  NBLKS       =          1280
>  NPROMA      =           128
>  Host allocate [ms]:             1
> 
> ### ITERATION 1/1
>      NUMPROC=1, NUMOMP=1, NGTOT=163840, NGPTOTG=163840, NPROMA=128, NGPBLKS=1280
>  Device allocate [ms]:            32
>  Reference MFLOP count for 100 columns :  12.48232900
>      NUMOMP    NGPTOT  #GP-cols     #BLKS    NPROMA tid# : Time(msec)  MFlops/s     col/s
>           1    163840         0         0       128    0 :         21         0   7784519 @ rank#0:core#62
>           1    163840         0      1280       128   -1 :         74    275665   2208443 : TOTAL @ rank#0
>     1 x   1    163840         0      1280       128   -1 :         74    275665   2208443 : TOTAL
> ```
> 
>  * [x]  Finally, after fixing the issues mentioned in the comments, the following test doesn't work for me right now on AC GPUs (both in single and double precision and with and without OpenMP):
> 
> ```
> 21/46 Test: dwarf-cloudsc-fortran-atlas-dp-serial
> Command: "/.../nab_atlas_gpu_field/build/bin/dwarf-cloudsc-fortran-atlas-dp" "1" "100" "16"
> Directory: /.../nab_atlas_gpu_field/build
> "dwarf-cloudsc-fortran-atlas-dp-serial" start time: May 26 06:09 CEST
> Output:
> ----------------------------------------------------------
>   LMULTIFIELD:   T
> (E) Exception: FieldSet: cannot find field "PLCRIT_AER"  (/.../nab_atlas_gpu_field/source/atlas/src/atlas/field/FieldSet.cc:120 field)
> 
> (E) =========================================
> (E) TERMINATING dwarf-cloudsc-fortran-atlas-dp
> (E) -----------------------------------------
> (E) FieldSet: cannot find field "PLCRIT_AER"
> (E) -----------------------------------------
> (E) LOCATION:  (/.../nab_atlas_gpu_field/source/atlas/src/atlas/field/FieldSet.cc:120 field)
> (E) -----------------------------------------
> (E) BACKTRACE
> (E) ------------------
> ```
> 
> And the SCC-HOIST test fails with a segfault:
> 
> ```
> 24/46 Test: dwarf-cloudsc-gpu-scc-hoist-atlas-dp
> Command: "/.../nab_atlas_gpu_field/build/bin/dwarf-cloudsc-gpu-scc-hoist-atlas-dp" "1" "1000" "128"
> Directory: /.../nab_atlas_gpu_field/build/cloudsc-dwarf/src/cloudsc_gpu_atlas.dp/../../..
> "dwarf-cloudsc-gpu-scc-hoist-atlas-dp" start time: May 26 06:09 CEST
> Output:
> ----------------------------------------------------------
>  LMULTIFIELD =   F
>  NGPTOTG     =          1000
>  NGPTOT      =          1000
>  NBLKS       =             8
>  NPROMA      =           128
>  LMULTIFIELD =   F
>  NGPTOTG     =          1000
>  NGPTOT      =          1000
>  NBLKS       =             8
>  NPROMA      =           128
> 
> ### ITERATION 1/1
>      NUMPROC=1, NUMOMP=1, NGTOT=1000, NGPTOTG=1000, NPROMA=128, NGPBLKS=8
> <end of output>
> Test time =   0.38 sec
> ----------------------------------------------------------
> Test Failed.
> ```

Thank you very much @reuterbal for having a thorough look on Atlas part of the dwarf! Much appreciated!

In addition to my previous answers, I can add here:
1) the per-thread throughput reporting is back again;
2) the CPU code version in the folder 'cloudsc-fortran-atlas' has been fixed;
3) the GPU code version in the folder 'clodusc-gpu-atlas' has been fixed to also run on CPU;
4) if cmake omits the folder 'prototype1' and one or two more, you can compile the Atlas part of the dwarf on GraceHopper as well. I have checked that the code executes on GH200 correctly.

So far, re Atlas seems to work in intended configurations, but I might have missed a configuration of two which we do not support through our tests. I will be happy to look into a specific problems should such needed to be fixed.

Again, many thanks for your invaluable comments!

sbrdar:
> > I would advise to reuse CPU and GPU codes where possible, to avoid code duplication and maintainability. In particular, I think the validate, and creation of the fields and loading the data is probably identical. The CPU OpenMP driver or GPU OpenACC driver can then be called independently.
> 
> I have now reused the expand_atlas_mod and validate_atlas_mod from 'cloudsc_fortran_atlas' in 'cloudsc_gpu_atlas'.

Many thanks for review this PR, @wdeconinck ! Very insightful hints and remarks! Thank you!

src/cloudsc_gpu_atlas/cloudsc_driver_scc_mod.F90
```diff
@@ -183,9 +184,12 @@ SUBROUTINE CLOUDSC_KERNEL( TIMER, NGPTOT, NGPBLKS, NLEV, NPROMA, PTSPHY, &
   call device_allocator%allocate(ZFOEELIQT, [NPROMA, NLEV, NGPBLKS])
 #endif
 
-!#if (ATLAS_HAVE_GPU == 0)
+! temporary disable floating-point-trapping of FE_INVALID caused by
+! too aggressive optimisation within the CLOUDSC function
+#if (ATLAS_HAVE_GPU == 0)
```
wdeconinck: I don't think the ATLAS_HAVE_GPU variable is defined at all. You should remove the #if, just keeping the IEEE_*_HALTING_MODE statements.

src/cloudsc_gpu_atlas/cloudsc_driver_scc_mod.F90
```diff
@@ -258,12 +262,12 @@ SUBROUTINE CLOUDSC_KERNEL( TIMER, NGPTOT, NGPBLKS, NLEV, NPROMA, PTSPHY, &
 !$acc end parallel loop
 !$acc end data
 
-!#if (ATLAS_HAVE_GPU == 0)
-    CALL IEEE_SET_HALTING_MODE(IEEE_INVALID, .TRUE.)
-!#endif
-
     CALL TIMER%THREAD_END(TID)
 
+#if (ATLAS_HAVE_GPU == 0)
```
wdeconinck: Also this #if should go, but keep what's nested.

wdeconinck (changes requested):

src/cloudsc_gpu_atlas/cloudsc_driver_scc_mod.F90
sbrdar: Thanks. Removed now.

src/cloudsc_gpu_atlas/cloudsc_driver_scc_mod.F90
sbrdar: .. and this one as well. Thanks.

wdeconinck:
linux-nvhpc ci fails with:
```
FAILED: [code=2] src/cloudsc_gpu_atlas.dp/CMakeFiles/dwarf-cloudsc-gpu-scc-atlas-dp.dir/cloudsc_driver_scc_mod.F90.o src/cloudsc_gpu_atlas.dp/module-dwarf-cloudsc-gpu-scc-atlas/cloudsc_driver_scc_mod.mod 
  /opt/nvhpc/Linux_x86_64/24.5/compilers/bin/nvfortran -I/home/runner/work/dwarf-p-cloudsc/dwarf-p-cloudsc/src/cloudsc_gpu_atlas -I/home/runner/work/dwarf-p-cloudsc/dwarf-p-cloudsc/src/common/include -I/home/runner/work/dwarf-p-cloudsc/dwarf-p-cloudsc/build/src/common.dp/module-atlas -I/home/runner/work/_temp/install/atlas/include -I/home/runner/work/_temp/install/atlas/module/atlas -I/home/runner/work/_temp/install/fckit/include -I/home/runner/work/_temp/install/fckit/module/fckit -I/home/runner/work/_temp/install/fckit/include/fckit -I/home/runner/work/_temp/deps/hdf5/include -Mlarge_arrays -g -O0 -Mbounds -traceback -module src/cloudsc_gpu_atlas.dp/module-dwarf-cloudsc-gpu-scc-atlas -fPIE -mp -c src/cloudsc_gpu_atlas.dp/CMakeFiles/dwarf-cloudsc-gpu-scc-atlas-dp.dir/cloudsc_driver_scc_mod.F90-pp.f90 -o src/cloudsc_gpu_atlas.dp/CMakeFiles/dwarf-cloudsc-gpu-scc-atlas-dp.dir/cloudsc_driver_scc_mod.F90.o
  NVFORTRAN-F-0004-Unable to open MODULE file pluto_module.mod (/home/runner/work/dwarf-p-cloudsc/dwarf-p-cloudsc/src/cloudsc_gpu_atlas/cloudsc_driver_scc_mod.F90: 30)
  NVFORTRAN/x86-64 Linux 24.5-1: compilation aborted
  ```
 Perhaps we need to link the target `dwarf-cloudsc-gpu-scc-atlas-${prec}` with `pluto_f` directly.

**Comment by wdeconinck** [comment]
The runner OS for intel-classic should be an older ubuntu-22.04. intel-llvm with ubuntu-24.04 should be OK.

**Comment by wdeconinck** [comment]
The ci for "linux gnu-14" is failing in atlas where OpenMP is seemingly detected by CMake with -fopenmp flag but then the C++ code cannot find the "omp.h" include file for some weird reason. To be investigated why...

**Comment by wdeconinck** [comment]
> Perhaps we need to link the target `dwarf-cloudsc-gpu-scc-atlas-${prec}` with `pluto_f` directly.

This was already the case, so the problem is in the export of pluto in install-interface.
This is fixed in atlas branch hotfix/0.43.1, to be released as 0.43.1 once these tests pass.

Please update in builds.yml "refs/tags/0.43.0" to "hotfix/0.43.1" to see. If all works, I will release 0.43.1 and then we can update the tag once more to "refs/tags/0.43.1"

**Comment by wdeconinck** [comment]
It seems the OpenMP for the gnu-14 on ubuntu-24.04 is really problematic. Perhaps we need to turn off OMP feature for this entire build?

intel-classic build shows that in case of multifield the expand is creating temporaries. This is because the dimensions of the second argument field have contiguous dimensions whereas the field passed is discontiguous slice. Changing the dimensions from `(nproma, nlev, nblk)` to `(:,:,:)` will solve this.

```
 43/48 Test #43: dwarf-cloudsc-gpu-scc-k-caching-multifield-atlas-sp ....***Failed    1.02 sec
   LMULTIFIELD =  T
   NGPTOTG     =         1000
   NGPTOT      =         1000
   NBLKS       =            8
   NPROMA      =          128
   Host allocate [ms]:       -56598
  forrtl: warning (406): fort: (1): In call to EXPAND_R2, an array temporary was created for argument #2
  
  Image              PC                Routine            Line        Source             
  dwarf-cloudsc-gpu  0000000000443103  expand_atlas_mod_          87  expand_atlas_mod.F90
  dwarf-cloudsc-gpu  000000000043847F  cloudsc_global_at         176  cloudsc_global_atlas_state_mod.F90
  dwarf-cloudsc-gpu  00000000004DA241  MAIN__                     78  dwarf_cloudsc_atlas.F90
  dwarf-cloudsc-gpu  000000000042B6CD  Unknown               Unknown  Unknown
  libc.so.6          00007FC53F229D90  Unknown               Unknown  Unknown
  libc.so.6          00007FC53F229E40  __libc_start_main     Unknown  Unknown
  dwarf-cloudsc-gpu  000000000042B5E5  Unknown               Unknown  Unknown
  ```

src/common/module/expand_mod.F90
```diff
@@ -234,11 +234,18 @@ subroutine expand_i1(buffer, field, nlon, nproma, ngptot, nblocks)
   end subroutine expand_i1
 
   subroutine expand_r1(buffer, field, nlon, nproma, ngptot, nblocks)
+    use omp_lib
```
wdeconinck: @sbrdar no need to add "use omp_lib" unless you are calling actual routines instead of pragmas.

src/cloudsc_fortran_atlas/cloudsc_global_atlas_state_mod.F90
```diff
@@ -243,78 +244,48 @@ SUBROUTINE CLOUDSC_GLOBAL_ATLAS_STATE_LOAD(SELF, FSET, FSPACE, NPROMA, NGPTOTG)
...
-      END DO
-      MULTIFIELD_IN = atlas_MultiField(CONFIG)
-      CALL FSET%ADD(MULTIFIELD_IN%FIELDSET()) 
+      MULTIFIELD_IN = atlas_MultiField(ATLAS_REAL(JPRB), [NPROMA, FSPACE%LEVELS(), -1, FSPACE%NBLKS()], IN_VAR_NAMES(1:24))
+      CALL FSET%ADD(MULTIFIELD_IN%FIELDSET())
       CALL MULTIFIELD_IN%FINAL()
     ELSE
-      DO IVAR = 1, INVAR_SIZE
-          CALL FSET%ADD(FSPACE%CREATE_FIELD(NAME=TRIM(IN_VAR_NAMES(IVAR)),   KIND=ATLAS_REAL(JPRB)))
+      DO IVAR = 1, 24
```
wdeconinck: I don't like the number 24 here. Can this be a more descriptive size variable?

src/cloudsc_fortran_atlas/cloudsc_global_atlas_state_mod.F90
```diff
@@ -243,78 +244,48 @@ SUBROUTINE CLOUDSC_GLOBAL_ATLAS_STATE_LOAD(SELF, FSET, FSPACE, NPROMA, NGPTOTG)
...
-        MULTIFIELD_OUT = atlas_MultiField(CONFIG)
-        CALL FSET%ADD(MULTIFIELD_OUT%FIELDSET()) 
-        CALL MULTIFIELD_OUT%FINAL()
+      MULTIFIELD_OUT = atlas_MultiField(ATLAS_REAL(JPRB), [NPROMA, FSPACE%LEVELS()+1, -1, FSPACE%NBLKS()], OUT_VAR_NAMES(1:14))
+      CALL FSET%ADD(MULTIFIELD_OUT%FIELDSET()) 
+      CALL MULTIFIELD_OUT%FINAL()
     ELSE
-      DO IVAR = 1, OUTVAR_SIZE
-          CALL FSET%ADD(FSPACE%CREATE_FIELD(NAME=TRIM(OUT_VAR_NAMES(IVAR)),   KIND=ATLAS_REAL(JPRB), LEVELS=SELF%KLEV+1))
+      DO IVAR = 1, 14
```
wdeconinck: I don't like the number 14 here. Can this be a more descriptive size variable?

src/cloudsc_gpu_atlas/CMakeLists.txt
```diff
@@ -0,0 +1,143 @@
+# (C) Copyright 1988- ECMWF.
```
wdeconinck: 1988 -> 2025

arch/ecmwf/gh/toolchain.cmake
```diff
@@ -0,0 +1,68 @@
+# (C) Copyright 1988- ECMWF.
```
wdeconinck: 1988 -> 2025

arch/ecmwf/gh/env.sh
```diff
@@ -0,0 +1,62 @@
+#! /usr/bin/env bash
+# (C) Copyright 1988- ECMWF.
```
wdeconinck: 1988 -> 2025

src/cloudsc_fortran_atlas/cloudsc_global_atlas_state_mod.F90
sbrdar: Good point. Thanks! We now use NUM_IN_MFIELDS and NUM_OUT_MFIELDS in cloudsc_fortran_atlas. Also, a similar change has been done in cloudsc_gpu_atlas.

src/cloudsc_fortran_atlas/cloudsc_global_atlas_state_mod.F90
sbrdar: Done. See previous comment.

src/cloudsc_gpu_atlas/CMakeLists.txt
sbrdar: Done.

arch/ecmwf/gh/toolchain.cmake
sbrdar: Done.

arch/ecmwf/gh/env.sh
sbrdar: Done. Thanks!

wdeconinck:
All CI is green now! I guess now it's time to clean up (rebase etc) the branch a bit before merging it.
Perhaps @reuterbal can have another review round to double check and approve.

.github/workflows/build-hpc.yml
```diff
@@ -46,14 +47,14 @@ jobs:
             sbatch_options: |
               #SBATCH --time=00:20:00
               #SBATCH --nodes=1
-              #SBATCH --ntasks=1
-              #SBATCH --cpus-per-task=32
+              #SBATCH --ntasks-per-node=4
+              #SBATCH --cpus-per-task=64
```
wdeconinck, line 50: This makes for speedy compilation, giving access to 256 threads on the hpc node.

.github/workflows/build-hpc.yml
```diff
@@ -76,7 +79,8 @@ jobs:
             sbatch_options: |
               #SBATCH --time=00:20:00
               #SBATCH --nodes=1
-              #SBATCH --ntasks-per-node=1
+              #SBATCH --ntasks-per-node=8
+              #SBATCH --cpus-per-task=7
```
wdeconinck, line 82: This gives access to 56 threads to compile on the dev-g LUMI node.
Empirically I found 56 is the maximum one can get.

.github/workflows/build-hpc.yml
```diff
@@ -8,6 +8,7 @@ on:
     branches:
     - main
     - develop
+    - 'nab_atlas_gpu_field'
```
wdeconinck: This should be removed before merging!

arch/ecmwf/gh/env.sh
```diff
@@ -0,0 +1,62 @@
...
+module use /opt/eb_packages/modules/all
+
+#module_load CMake/3.27.6-NVHPC-24.1-CUDA-12.3.0
+module_load Python/3.11.3-GCCcore-12.3.0
+module_load OpenMPI/4.1.6-GCC-13.2.0
+module_load NVHPC/24.1-CUDA-12.3.0
+module_load HDF5/1.14.3-NVHPC-24.1-CUDA-12.3.0
+module_load Boost/1.82.0-GCC-12.3.0
+
+export PATH=/perm/nawd/software/gh/cmake/3.30.3/bin:$PATH
```
wdeconinck, line 45: This needs a more permanent solution!

.github/workflows/build-hpc.yml
reuterbal: Yes, that's correct - see the disclaimer here: https://docs.lumi-supercomputer.eu/runjobs/scheduled-jobs/lumig-job/

reuterbal (approved):
Many thanks, to me this looks very good! The TLC applied to the build-hpc workflow is also much appreciated

.github/workflows/build-hpc.yml
sbrdar: Done.

wdeconinck (approved):

PR #90, opened 2024-05-31
CLOUDSC_FIELD_STATE_MOD: fix device memory leak for packed storage option

awnawab:
A small PR to fix a device memory leak if the `CLOUDSC_PACKED_STORAGE` option is enabled for the scc-field variant.


reuterbal (approved):

mlange05 (approved):
Yes, that looks sensible. GTG from me.

PR #92, opened 2024-07-17
Single precision and architecture files for SYCL variants

MichaelSt98:
SYCL SP tested on AC

via e.g. `cloudsc-bundle build --clean --build-dir=build-sp-sycl-serialbox --arch=arch/ecmwf/hpc2020/intel-sycl/2021.4.0 --with-sycl --with-serialbox --single-precision --cloudsc-cpp-math=C` ...


bundle.yml
```diff
@@ -123,6 +124,14 @@ options :
         cmake: >
             ENABLE_SYCL=ON
 
+    - sycl-sub-group-size :
```
reuterbal, line 127: The SYCL variants don't like non-matching NGPTOT, NPROMA and SUB_GROUP_SIZE values, right? I'm seeing errors like:
```
$ bin/dwarf-cloudsc-scc-sycl 1 100 32
Running on NVIDIA A100-SXM4-40GB
terminate called after throwing an instance of 'sycl::_V1::nd_range_error'
  what():  Non-uniform work-groups are not supported by the target device -54 (PI_ERROR_INVALID_WORK_GROUP_SIZE)
Aborted
```

This affects, in particular, also CTest where we (intentionally) use a non-power-of-2 NGPTOT. So, I suggest to add a check after reading the arguments and exiting gracefully, and changing the CTest definition to a power of 2 for NGPTOT.

reuterbal (changes requested):
Thanks, looks very good and confirmed working with offloading on AC.

I've left a few minor comments, and it would be good to fail gracefully for unsupported test case dimensions (and change the ctest so that it is valid).

bundle.yml
MichaelSt98: I've added a test to check whether NGPTOT and NPROMA match. I haven't added a check for the SUB_GROUP_SIZE as I expect people changing the default value to know what they are doing ... Is that ok or do you want to have another check?

src/cloudsc_sycl/cloudsc/dtype.h
```diff
@@ -0,0 +1,100 @@
...
+// FLOATING POINT PRECISION
+#ifdef SINGLE
+typedef float dtype;
+#else
+typedef double dtype;
+#endif
+
+
+// MATH FUNCTIONS
+#ifdef SINGLE
```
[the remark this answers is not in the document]
MichaelSt98: That is true, however, I would like to keep the current order (and duplication) as it is more clear and I could imagine that changes are necessary to this whenever we target new hardware ... 
However, if you insist I'll change it.

bundle.yml
reuterbal: That sounds good and agreed - as opposed to NGPTOT and NPROMA, the SUB_GROUP_SIZE isn't something that's exposed as prominently to the user.

reuterbal (approved):
Thanks for taking care of the requested changes! Looks all good to me now and confirmed to work locally.

PR #96, opened 2024-10-01
Updated Field API variants of the clouds dwarf (CPU and GPU)

wertysas:
**This is an update of the Field API variants of cloudsc that aims to resemble the use of Field API in the IFS more closely than before.** This includes the addition of a new global state type for the Field API variants: `CLOUDSC_FIELD_STATE_TYPE` and three types for aux vars, fluxes and tendencies that resemble similar types in the IFS:

- `CLOUDSC_AUX_TYPE`
- `CLOUDSC_FLUX_TYPE`
- `CLOUDSC_STATE_TYPE`

**There has also been updates to the build structure, that makes it possible to build the CPU Field API version without OpenACC or CUDA and to build the GPU Field API version without CUDA support**. The following ecbuild_options have been added to support this:

- `with-field-api` enables field-api variants to be built
- `with-mapped-fields` builds field-api variants with field pointers automatically by OpenACC to device (requires OpenACC and CUDA)
- `with-pinned-fields` builds field-api variants with fields allocated in pinned memory (requires OpenACC and CUDA)


FussyDuck:
[CLA assistant check](https://bol-claassistant.ecmwf.int:443/ecmwf-ifs/dwarf-p-cloudsc?pullRequest=96) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you all sign our [Contributor License Agreement](https://bol-claassistant.ecmwf.int:443/ecmwf-ifs/dwarf-p-cloudsc?pullRequest=96) before we can accept your contribution.<br/>**0** out of **2** committers have signed the CLA.<br/><br/>:x: Johan Ericsson<br/>:x: wertysas<br/><hr/>**Johan Ericsson** seems not to be a GitHub user. You need a GitHub account to be able to sign the CLA. If you have already a GitHub account, please [add the email address used for this commit to your account](https://help.github.com/articles/why-are-my-commits-linked-to-the-wrong-user/#commits-are-not-linked-to-any-user).<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://bol-claassistant.ecmwf.int:443/check/ecmwf-ifs/dwarf-p-cloudsc?pullRequest=96) it.</sub>

reuterbal (changes requested):
Many thanks, this looks like it was a substantial piece of work that has been carried out very diligently and well structured!

I left a few small comments but nothing major.

To make sure this is tested in CI, please add `--with-field-api` in all four entries in the github workflow here:
https://github.com/ecmwf-ifs/dwarf-p-cloudsc/blob/57e95e83be7ed3fe6e1fbabed95d915c231beaee/.github/workflows/build.yml#L35-L38

Further down, you may have to add the GPU-enabled field_api build target also to the ctest_exclude options, e.g., here: 
https://github.com/ecmwf-ifs/dwarf-p-cloudsc/blob/57e95e83be7ed3fe6e1fbabed95d915c231beaee/.github/workflows/build.yml#L57

In addition, you will have to add the new target to the script here: https://github.com/ecmwf-ifs/dwarf-p-cloudsc/blob/develop/.github/scripts/verify-targets.sh

reuterbal:
Apologies, final request: Please update also the README section here:
https://github.com/ecmwf-ifs/dwarf-p-cloudsc/blob/57e95e83be7ed3fe6e1fbabed95d915c231beaee/README.md?plain=1#L81-L90

src/cloudsc_gpu/CMakeLists.txt
```diff
@@ -258,7 +258,7 @@ if( HAVE_CLOUDSC_GPU_SCC_FIELD )
     )
 
     # Importantly, we add the "pinned" flag to the linker command to ensure pinning!
-    if( CMAKE_Fortran_COMPILER_ID MATCHES "NVHPC" )
+    if( (CMAKE_Fortran_COMPILER_ID MATCHES "NVHPC") AND field_api_HAVE_CUDA )
         target_link_options( dwarf-cloudsc-gpu-scc-field PUBLIC "-gpu=pinned" )
     endif()
```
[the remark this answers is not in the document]
awnawab: No sadly the pinning flags are not exported by F-API, just the acc and cuda flags. Once I've fixed this in F-API I'll remove this flag here in a future PR.

src/common/module/cloudsc_aux_type_mod.F90
```diff
@@ -0,0 +1,259 @@
...
+    REAL(KIND=JPRB), POINTER, CONTIGUOUS :: PCLV(:,:,:)
+    REAL(KIND=JPRB), POINTER, CONTIGUOUS :: PSUPSAT(:,:)
+    ! Output fields used for validation
+    REAL(KIND=JPRB), POINTER, CONTIGUOUS :: PCOVPTOT(:,:) ! Precip fraction
+    REAL(KIND=JPRB), POINTER, CONTIGUOUS :: PRAINFRAC_TOPRFZ(:)
+    
+    ! field gang for 3D fields
+    CLASS(FIELD_4RB), POINTER :: DATA_RDONLY
+    ! Acces pointers for 3D fields in gang
+    TYPE(FIELD_3RB_PTR), ALLOCATABLE :: FIELDS_RDONLY(:)
```
awnawab, line 62: This should never be accessed by external code, so it should be marked `PRIVATE`.

src/common/module/cloudsc_flux_type_mod.F90
```diff
@@ -0,0 +1,171 @@
...
+    REAL(KIND=JPRB), POINTER, CONTIGUOUS :: PFPLSL(:,:)   ! liq+rain sedim flux
+    REAL(KIND=JPRB), POINTER, CONTIGUOUS :: PFPLSN(:,:)   ! ice+snow sedim flux
+    REAL(KIND=JPRB), POINTER, CONTIGUOUS :: PFHPSL(:,:)   ! Enthalpy flux for liq
+    REAL(KIND=JPRB), POINTER, CONTIGUOUS :: PFHPSN(:,:)   ! Enthalp flux for ice
+    
+    CLASS(FIELD_3RB), POINTER :: F_PFSQLF, F_PFSQIF, F_PFCQLNG, F_PFCQNNG, F_PFSQRF, F_PFSQSF, &
+      & F_PFCQRNG, F_PFCQSNG, F_PFSQLTUR, F_PFSQITUR, F_PFPLSL, F_PFPLSN, F_PFHPSL, F_PFHPSN
+  
+    CLASS(FIELD_4RB), POINTER :: DATA_WRONLY ! ACTUAL FIELD storing data
+    TYPE(FIELD_3RB_PTR), ALLOCATABLE :: FIELDS_WRONLY(:) ! Array of field pointers
```
awnawab, line 42: Same comment here about privatization.

src/common/module/cloudsc_state_type_mod.F90
```diff
@@ -0,0 +1,110 @@
...
+      LOGICAL :: PACKED
+      REAL(KIND=JPRB), DIMENSION(:,:), POINTER :: T   ! GMV FIELDS
+      REAL(KIND=JPRB), DIMENSION(:,:), POINTER :: Q, A  ! GFL FIELDS
+      REAL(KIND=JPRB), DIMENSION(:,:,:), POINTER :: CLD   ! COMPOSED CLOUD ARRAY
+ 
+      CLASS(FIELD_3RB), POINTER :: F_T, F_A, F_Q
+      CLASS(FIELD_4RB), POINTER :: F_CLD
+
+      CLASS(FIELD_4RB), POINTER :: FIELD_GANG
+      TYPE(FIELD_3RB_PTR), PRIVATE, ALLOCATABLE :: FIELD_PTRS(:)
```
awnawab, line 32: Same comment here about privatization.

src/cloudsc_gpu/cloudsc_driver_gpu_scc_field_mod.F90
```diff
@@ -104,68 +104,78 @@ SUBROUTINE CLOUDSC_DRIVER_GPU_SCC_FIELD( &
...
+    CALL FIELD_STATE%AUX%F_PCCN%GET_DEVICE_DATA_RDONLY(PCCN)
+    CALL FIELD_STATE%AUX%F_PNICE%GET_DEVICE_DATA_RDONLY(PNICE)
+    
+    CALL FIELD_STATE%AUX%F_PAPH%GET_DEVICE_DATA_RDONLY(PAPH)
+    CALL FIELD_STATE%AUX%F_PLSM%GET_DEVICE_DATA_RDONLY(PLSM)
+    CALL FIELD_STATE%AUX%F_LDCUM%GET_DEVICE_DATA_RDONLY(LDCUM)
+    CALL FIELD_STATE%AUX%F_KTYPE%GET_DEVICE_DATA_RDONLY(KTYPE)
+    CALL FIELD_STATE%AUX%F_PCLV%GET_DEVICE_DATA_RDONLY(PCLV)
+    
     CALL FIELD_STATE%TENDENCY_TMP%F_T%GET_DEVICE_DATA_RDONLY(TEND_TMP_T)
```
awnawab, line 143: Should these not also get data movement calls on the FieldGang if the packed option is enabled, like we have for `FIELD_STATE%AUX` and `FIELD_STATE%FLUX`?

src/cloudsc_gpu/cloudsc_driver_gpu_scc_field_mod.F90
```diff
@@ -222,28 +232,33 @@ SUBROUTINE CLOUDSC_DRIVER_GPU_SCC_FIELD( &
     CALL TIMER%THREAD_END(TID)
 
     IF(USE_PACKED)THEN
-       CALL FIELD_STATE%DATA_RWONLY%SYNC_HOST_RDWR()
+       CALL FIELD_STATE%AUX%SYNC_HOST()
```
awnawab, line 235: The pullback-to-host should mirror the copy-to-device logic; we should call `SYNC_HOST_RDWR` on the FieldGang if `USE_PACKED` is true, and then simply call `SYNC_HOST_RDWR` on all the named children. Once `SYNC_HOST_RDWR` is called on the FieldGang, calling it on the children doesn't trigger any data movement. The same is also true of calling it twice on the same field (`GET_DEVICE_RDONLY` followed by `SYNC_HOST_RDWR` will also not move any data back to host).

I believe this change should make the `SYNC_HOST` methods redundant, so you could then remove those.

bundle.yml
```diff
@@ -92,9 +92,26 @@ options :
...
+        cmake : >
+            BUILD_field_api=ON
+            ENABLE_CLOUDSC_FORTRAN_FIELD=ON
+
+    - with-mapped-fields :
+        help : Enable Field API to use mapped memory by default
+        cmake : >
+            ENABLE_FIELD_API_MAPPED_MEMORY=ON
+    
+    - with-pinned-fields :
```
awnawab, line 106: This bundle option should then be removed once pinning is made run-time togglable.

awnawab (changes requested):
Many thanks @wertysas for this excellent and comprehensive piece of work! 🙏  Mirroring the F-API usage in the IFS is really great as it will allow us to test F-API related Loki recipes on cloudsc first.

I've left a few comments for you to address, but none of these require major changes.

src/cloudsc_gpu/cloudsc_driver_gpu_scc_field_mod.F90
wertysas: Good catch!

src/common/module/cloudsc_aux_type_mod.F90
wertysas: Currently I am calling the sync_device_rdonly() of the field gang from the driver, and this would have to be changed if I made this private. I can add a method to the aux, flux, and tendency objects to expose the sync method of the field gang in the driver if it is preferable. What do you think? @awnawab

src/cloudsc_loki/CMakeLists.txt
wertysas, line 1: This should probably not be added here, since this is the SCC CUF variant that is not based on the field variant of the gpu driver. My thinking is that a new loki_transform will be added through the dwarf-p-cloudsc/e-field-api-offload branch that I am using for development of the Loki Field API offload transformation.

src/cloudsc_loki/CMakeLists.txt
```diff
@@ -619,6 +619,10 @@ if( HAVE_CUDA )
             ${COMMON_MODULE}/yomphyder.F90
             ${COMMON_MODULE}/yoethf.F90
             ${COMMON_MODULE}/yoecldp.F90
+            ${COMMON_MODULE}/cloudsc_field_state_mod.F90
+            ${COMMON_MODULE}/cloudsc_flux_type_mod.F90
+            ${COMMON_MODULE}/cloudsc_aux_type_mod.F90
+            ${COMMON_MODULE}/cloudsc_state_type_mod.F90
```
wertysas, line 625: This should probably not be added here, since this is the SCC CUF variant that is not based on the field variant of the gpu driver. My thinking is that a new loki_transform will be added through the dwarf-p-cloudsc/je-field-api-offload branch that I am using for development of the Loki Field API offload transformation.

src/common/module/cloudsc_aux_type_mod.F90
awnawab: Sorry I realise now my comment was quite ambiguous. Whilst the FieldGang itself should remain publicly accessible, `FIELDS_RDONLY` which is used only to store pointers to the gang's children, should be made private. There shouldn't be a need to call this from external code.

src/cloudsc_gpu/cloudsc_driver_gpu_scc_field_mod.F90
wertysas: I don't really understand what difference is between SYNC_HOST and SYNC_HOST_RDWR for the FieldGang. Is it exactly the same as for normal fields? Might have to update this then.

awnawab: `SYNC_HOST` is not a F-API call but rather a type-bound method of the new derived-types you have introduced. I think a `SYNC_HOST` method exists for the corresponding types in ecphys. It makes a lot of sense when adding the F-API calls by hand but if the F-API boilerplate is to be loki generated then we shouldn't add data movement methods to IFS derived types and just call `SYNC_HOST_RDWR/RDONLY` on all the constituent fields.

Calling `SYNC/GET_HOST/DEVICE_*` methods on the FieldGang is functionality equivalent to calling the same methods one-by-one on all of the children, except that we will now do one contiguous copy rather than many strided copies.

awnawab:
Many thanks @wertysas for addressing all of my comments! 🙏 The CI fails at the moment because it can't find the `scc-field` targets. This is because the condition for that is `field_api_HAVE_ACC` which we explicitly disable. Changing that condition to `HAVE_ACC` should fix it. You might also want to set the env variable `DEV_ALLOC_SIZE=1073741824` (1Gb) for the F-API gpu enabled unit tests.

awnawab (approved):
Thanks a lot for sorting out the CI issues, this looks great and good to go 👌

reuterbal (changes requested):
Many thanks for the updates and @awnawab for the diligent review! Just one final request to use a more meaningful option description, otherwise ready to go!

CMakeLists.txt
```diff
@@ -111,12 +111,20 @@ endif()
...
     DESCRIPTION "Use field_api to manage GPU data offload and copyback"
     REQUIRED_PACKAGES "field_api"
-    CONDITION HAVE_CUDA
     DEFAULT ON )
 
 ecbuild_find_package( NAME loki )
 ecbuild_find_package( NAME atlas )
 
+ecbuild_add_option( FEATURE FIELD_API_DISABLE_MAPPED_MEMORY
+  DESCRIPTION "Use ACC mapped memory by default in Field API objects"
```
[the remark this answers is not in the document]
wertysas: Yes, thank you for spotting that, I missed to update the description when changing this option. I have changed it now.

mlange05 (approved):
Looks fantastic! Great contribution and enables a lot of great follow-on stuff. 

Driver structure and data types look much closer to the full system in this and the structural issue are solved quite elegantly (eg. the packed FIELD_GANG implementation. 

GTG from me. :shipit:

reuterbal (approved):

PR #99, opened 2024-11-04
loki_transform for CUF variants, bundle.yml fixup

reuterbal:
A small fixup to PR #94:
- The help string in bundle.yml made this invalid YAML, which broke the `--help` option of ecbundle
- the CUDA Fortran still used the outdated loki_transform_convert interface. I flipped this to loki_transform, which means we'll be able to remove the old macro from loki, too.


mlange05 (approved):
Yes, looks good to me.

PR #100, opened 2024-11-21
New Field API Loki variant with state types in driver

wertysas:
This PR adds a new Field API Loki variant , ``dwarf-cloudsc-loki-scc-field``,  that applies Loki transformations to a driver with fields that are grouped and wrapped by derived types. The main changes are

-  A new driver with loki offload annotations is added that is based on the Field API CPU variant from the recently merged PR: [Updated Field API variants of the clouds dwarf (CPU and GPU) #96](https://github.com/ecmwf-ifs/dwarf-p-cloudsc/pull/96)

-  A new config file with a pipeline ``scc-field`` is added that uses the ``FieldOffloadTransformation`` from the recently merged Loki PR: [DataOffload: Add new FieldOffloadTransformation for FIELD API boilerplate injection #437](https://github.com/ecmwf-ifs/loki/pull/437)

And some CMake and bundle boilerplate. **Note that the Loki version is changed from ``v.0.2.7`` to ``v0.2.9``**


src/cloudsc_loki/cloudsc_driver_field_loki_mod.F90
```diff
@@ -0,0 +1,140 @@
...
+      !$omp end do nowait
+
+      CALL TIMER%THREAD_END(TID)
+
+      !$omp end parallel
+
+      !$loki end data
+
+      CALL TIMER%END()
+
```
reuterbal, line 133: Please add here the following block 
```suggestion

#ifdef CLOUDSC_GPU_TIMING
    ! On GPUs, adding block-level column totals is cumbersome and
    ! error prone, and of little value due to the large number of
    ! processing "thread teams". Instead we register the total here.
    CALL TIMER % THREAD_LOG(TID=TID, IGPC=NGPTOT)
#endif
```
Otherwise the throughput performance reporting is broken when running on GPU

as in the other loki_driver_mod https://github.com/ecmwf-ifs/dwarf-p-cloudsc/blob/82fdf4b4021b37962c3c129c7b0047d72c39ee20/src/cloudsc_loki/cloudsc_driver_loki_mod.F90#L182-L187

reuterbal (changes requested):
Many thanks, this is fantastic! I can confirm that it works on A100, produces correct results and expected performance.

One small request to fix the performance reporting, otherwise good to go!

reuterbal (approved):
Thanks! Confirmed working and performance reporting looks good.

mlange05 (approved):
Looks fantastic! What a great PR to get the century PR tag and end the year with. Many thanks to @wertysas and @reuterbal for final review and integration. GTG from me. :shipit:

PR #105, opened 2024-12-19
Loki: Always inline stmt functions from C-headers

mlange05:
This PR removes the use of the re-written one-line elemental functions that where used to facilitate inlining of the historical "commdeck" functions. This was a hard requirement for all flavours of C-transpilation. However, Loki can now also inline raw statement functions, if they have been internalised to the kernel via preprocessing. In this PR we now switch this on in the Loki config and remove the special treatment from the CMake config.

Note, I've also piggy-backed an update to the latest Loki version and the removal or a rogue import, that can cause problems.

Another note, I've also left the module-based elemental functions in the source, for reference, and because there are CUF-specific versions in the same place. I've only cleaned up the raw C-transpilation, but not yet any of CUF modes.


mlange05:
Integration note: This requires a very recent Loki commit, which has not been tagged yet: 
https://github.com/ecmwf-ifs/loki/commit/ad1e2bc76f342a59ae2edb383ae2393af2d111bb

MichaelSt98 (approved):
Nice! I agree, there is no need to keep the re-written one-line elemental functions.

src/cloudsc_loki/CMakeLists.txt
```diff
@@ -30,12 +30,6 @@ if( HAVE_CLOUDSC_LOKI )
 
     set( LOKI_FRONTEND "fp" CACHE STRING "Frontend parser for Loki transforms" )
 
-    # OFP frontend cannot deal with statement functions, so we toggle them here
-    set( CLOUDSC_DEFINE_STMT_FUNC "" )
-    if(NOT "${LOKI_FRONTEND}" STREQUAL "ofp")
-    set( CLOUDSC_DEFINE_STMT_FUNC CLOUDSC_STMT_FUNC )
-    endif()
-
```
reuterbal, line 38: :heart:

reuterbal (approved):
Very nice! Are the `fcttre_mod.F90` and `fccld_mod.F90` files still used anywhere? Otherwise I would suggest removing them.

reuterbal:
Oh, the Loki upgrade in #100 has incurred a conflict, this branch needs a rebase now

mlange05:
> Very nice! Are the `fcttre_mod.F90` and `fccld_mod.F90` files still used anywhere? Otherwise I would suggest removing them.

No they are not used anymore. Their equivalents in CUF-land are tho, so I left them in. I'll remove the original now, we can remove the CUD ones once the pipeline is adjusted there (sorry, did this one quickly).

PR #106, opened 2024-12-24
Loki: C transpilation from pipeline in config file

mlange05:
This PR adds a new C-transpilation pipeline to the Loki configuration file that replaces the previous use of the implicit entry point for the `FortranCTransformation` in `loki-transform.py convert`. It does this by adding a new entry for the `FortranISOCWrapperTransformation` and marking the corresponding modules with `"header"` roles.

Integration note: This PR requires Loki PR #464 (https://github.com/ecmwf-ifs/loki/pull/464) to be merged first, and currently points to this branch in a temporary head commit that should be removed before merging.


reuterbal (approved):
Looks good to me :shipit:

PR #107, opened 2025-01-06
Loki low-level GPU via Loki (CUDA C/Fortran)

MichaelSt98:
* (re-)enable CUDA C Loki transpilation
* make custom/additional entry points/source code for CUDA C/Fortran transpilation obsolete


MichaelSt98:
@reuterbal and @mlange05 what do you think about disabling the failing tests for now (single-precision not yet supported for CUDA transpilation and there seems to be a serialbox/atlas problem?!) and fix that with additional/future PRs?

PR #110, opened 2025-01-17
CLOUDSC: Purge Loki-CLAW and GPU-CLAW variants from dwarf

mlange05:
o7


mlange05:
Note, this now sits on top of #111 to get the test base green. It thus also depends on an open Loki PR and has a hidden re-pointing commit that needs removing.

reuterbal (approved):
Thanks, another one bites the dust. I appreciate all the careful updating of the README and testing! GTG

PR #111, opened 2025-01-18
Loki: Small fixes and clean-up; EC_PMON removal

mlange05:
Following recent upstream changes in Loki the test case was no longer clean. This PR brings a few small changes that should remedy that, and removes the use of EC_PMON, an outdated power measurement utility on the old Cray system.

In more detail:
* Removal of deprecated / removed CMake options for Loki + CUDA variants
* Latest FIELD API version to silence GNU compiler warnings
* Properly guard module imports in Loki + Field API offload variants
* Updated CMake modules for ECMWF GNU toolchains
* Removal of EC_PMON from the Fortran and Field-based Fortran drivers


reuterbal (approved):
Thanks, the changes look good. The Loki fix has been merged, feel free to redirect Loki back to main and merge this.

PR #13, opened 2022-04-26
Fix compilation on macOS

wdeconinck:

PR #11, opened 2022-04-21
NVHPC HPC2020 arch

reuterbal:
Filing this early, feel free to play around and suggest changes when you find the time.

This time, README is updated accordingly and should contain all relevant information on how to build and run.

PR #10, opened 2022-03-09
HPC2020 architecture files

reuterbal:
This adds arch files for GNU 9.3.0, GNU 11.2.0 and Intel 2021.4.0 on Atos HPC, and removes the obsolete TEMS architecture files.

Note:
- GNU 11 builds with MPI enabled fail without `-fallow-argument-mismatch`. This is terrible but might require switching from `#include mpif.h` to `use mpi_f08`.
- CLAW/OMNI does not seem to compile with GNU 11 (failing on duplicate symbols somehow)

Very rough performance results, with GNU11 doing surprisingly well (using NPROMA=64 for single precision runs):
```
export OMP_NUM_THREADS=64
OMP_PLACES="{$(seq -s '},{' 0 $(($OMP_NUM_THREADS-1)) )}" srun -q np --ntasks=1 --hint=nomultithread --cpus-per-task=$OMP_NUM_THREADS ./bin/dwarf-cloudsc-fortran $OMP_NUM_THREADS 163840 32

GNU 11.2.0 
- dp: 73.252 GFlop/s
- sp: 96.214 GFlop/s

GNU 9.3.0
- dp: 66.945 GFlop/s
- sp: 87.330 GFlop/s

Intel 2021.4.0
- dp: 66.866 GFlop/s
- sp: 77.242 GFlop/s

----

export OMP_NUM_THREADS=8
OMP_PLACES="cores" srun -q np --ntasks=$(( 64 / $OMP_NUM_THREADS )) --hint=nomultithread --cpus-per-task=$OMP_NUM_THREADS ./bin/dwarf-cloudsc-fortran $OMP_NUM_THREADS 163840 32

GNU 11.2.0 
- dp: 73.961 GFlop/s
- sp: 97.464 GFlop/s

GNU 9.3.0
- dp: 70.608 GFlop/s
- sp: 89.142 GFlop/s

Intel 2021.4.0
- dp: 70.476 GFlop/s
- sp: 79.282 GFlop/s
```

Deviations from Leap42/GNU 7.3.0 reference results are negligible for GNU builds (CLOUDSC's validation module doesn't even flag them with `!!!!`) and reasonably small for Intel builds.

PR #113, opened 2025-02-14
Refactor CMake build scripts

reuterbal:
This is a substantial amount of changes to facilitate a cleaner, more modular build system for the dwarf. This aligns this with other repos and allows to update the CI testing to the current standard via ecmwf actions.

I fully expect that this still has some problems and requires a few changes before it can be integrated, but I would appreciate if you could take this for a spin and point out any problems that you may be seeing on various systems. To split the load, I would appreciate feedback from:

- [x] @mlange05 on the cloudsc_fortran, cloudsc_c, cloudsc_loki, cloudsc_python, cloudsc_pyiface variants
- [x] @MichaelSt98 on the cloudsc_gpu, cloudsc_hip, cloudsc_cuda, cloudsc_sycl variants
- [x] @wdeconinck on the cloudsc_atlas variant (and any pointers regarding the CMake, obviously)
- [x] @awnawab and @wertysas on the field-api variants

From an outward facing point of view, this has the following changes:

- Not every flavour is treated as a separate feature anymore - instead I am capturing the programming models or dependencies as features and always build everything for which requirements are fulfilled. Only exceptions to this are
  - the MPI support, which is a bit artificial anyway and remains an option that should be enabled explicitly
  - the Python variants
- To align this, I have also cleaned up the bundle.yml and converted all options to the form `--with-<feature>=[ON|OFF]`. Loki, Field-API and Atlas are not enabled by default to keep the vanilla invocation slim. This allows to explicitly request certain features and therefore let the bundle build fail if it cannot be enabled, while the default will gracefully step over it. Example of this effect for CUDA support:
  - On a default Linux system with GNU compiler and HDF5 installed, `./cloudsc-bundle build` will not build cuda variants
  - On the same system, building with `./cloudsc-bundle build --with-cuda=ON` will fail because it won't be able to fulfill the CUDA requirement
  - On our GPU partition with NVHPC, `./cloudsc-bundle build` will compile the CUDA variant by default without the need to specify the feature explicitly

Under the hood are a few more changes:

- All binaries are now precision-specific and by default, both single and double precision variants are built at the same time
- Instead of just one common lib we now compile several, programming model-specific versions with separate module directories. This allows to compile for example OpenMP and OpenACC variants without them accidentally interacting with each other. As a bonus, we don't have any accidental aliasing anymore and Ninja builds are working.
- Loki variants now use the `loki_transform_target` macro, except for the C transpilation
- Testing across several systems and compilers showed several outdated/defunc arch files, which I removed
- I have also made an effort to clean up several compiler warnings that were showing up everywhere.

A few problems I encountered an where I would like to get some feedback (mostly questions to @MichaelSt98):

- ~SYCL variants seem to be incompatible with newer Intel compilers, particularly the GPU_MATH options don't all work anymore and some API changes seem to prevent compilation~
- ~The HIP variants mostly fail with `Memory access fault by GPU node-4 (Agent handle: 0x390ac0) on address (nil). Reason: Unknown.` except for the hoist and k-caching variants in single precision. Which suggests to me that the build procedure seems ok in principle but may be missing a tiny detail that I'm unable to figure out?~
- ~The OpenMP GPU variants seem to be producing wrong results~

Given how extensive these changes are, I would suggest that when this PR is ready, that we maybe tag the current state of develop as a release before integration.

Next steps will then be:
- Extend CI to Atlas and Field-API variants
- Extend CI to build-hpc workflow on AC
- Extend CI to build-hpc workflow on LUMI-G

*Edit:* Thanks to @MichaelSt98 's help, the listed problems have been resolved.


awnawab (approved):
Thanks a lot for this mammoth cleanup. The new cmake setup is much nicer and easier to work with indeed, and the all F-API related changes look good 👌 🙏

arch/ecmwf/hpc2020/intel-sycl/2021.4.0/env.sh
```diff
@@ -39,12 +32,11 @@ module_load hdf5/1.10.6
 module_load cmake/3.20.2
```
MichaelSt98, line 32: I had problems with `cmake/3.20.2` and used instead `cmake/3.28.3`

MichaelSt98 (approved):
Thank you very much @reuterbal. This is awesome!

reuterbal:
Thanks for the reviews so far!

A nice side-effect of the refactoring is that we have a more structured approach to testing, and it seems that for most variants we are even fine to run the GPU code on the host for the Github Actions. Notable exceptions are of course CUDA and CUDA Fortran, and some OpenMP-target variants seem to segfault or have other problems while running fine on GPU. I'm currently excluding them from the ctest execution and would suggest we add coverage for those via the planned build-hpc workflow instead.

I have also tried to add the Field-API variants to the Github Action but these fail for almost all compilers:
- GNU 14: Test fails with SIGABORT
```
  Program received signal SIGABRT: Process abort signal.
  
  Backtrace for this error:
  #0  0x7f3dc8223e59 in ???
  #1  0x7f3dc8222e75 in ???
  #2  0x7f3dc7e4532f in ???
  #3  0x7f3dc7e9eb2c in pthread_kill
  #4  0x7f3dc7e4527d in gsignal
  #5  0x7f3dc7e288fe in abort
  #6  0x7f3dc85cc0c4 in dev_malloc.cold
  #7  0x7f3dc86c4a0b in __dev_alloc_module_MOD_field_3rm_dev_allocate_dim
  #8  0x7f3dc86c4a96 in __dev_alloc_module_MOD_field_3rm_dev_allocate_hst
  #9  0x7f3dc8623fc1 in __field_3rm_module_MOD_field_3rm_get_device_data
  #10  0x7f3dc8623b0f in __field_3rm_module_MOD_field_3rm_owner_get_device_data
  #11  0x55f45d8951fa in __cloudsc_driver_field_loki_mod_MOD_cloudsc_driver_field
  #12  0x55f45d8946e2 in MAIN__
  #13  0x55f45d89445e in main
  ```
- GNU 13 with MPI: similar symptom
```
   Program received signal SIGABRT: Process abort signal.
  
  Backtrace for this error:
  #0  0x7f5f91223e59 in ???
  #1  0x7f5f91222e75 in ???
  #2  0x7f5f90c4532f in ???
  #3  0x7f5f90c9eb2c in pthread_kill
  #4  0x7f5f90c4527d in gsignal
  #5  0x7f5f90c288fe in abort
  #6  0x7f5f90fad0b4 in dev_malloc.cold
  #7  0x7f5f9107037f in __dev_alloc_module_MOD_field_3rm_dev_allocate_dim
  #8  0x7f5f910703ff in __dev_alloc_module_MOD_field_3rm_dev_allocate_hst
  #9  0x7f5f90ff16e9 in __field_3rm_module_MOD_field_3rm_get_device_data
  #10  0x7f5f90ff1310 in __field_3rm_module_MOD_field_3rm_owner_get_device_data
  #11  0x56069f5da23f in __cloudsc_driver_field_loki_mod_MOD_cloudsc_driver_field
  #12  0x56069f5d972b in MAIN__
  #13  0x56069f5d94ae in main
  ```
- Intel Classic: fiat doesn't compile, with loads of errors like this:
```
  In file included from /usr/include/c++/13/cstdlib(81),
                   from /usr/include/c++/13/stdlib.h(36),
                   from /home/runner/work/_temp/download/fiat/src/fiat/system/internal/cxxdemangle.cc(12):
  /usr/include/c++/13/bits/std_abs.h(142): error: identifier "__builtin_fabsf128" is undefined
        return __builtin_fabsf128(__x);
               ^
  
  compilation aborted for /home/runner/work/_temp/download/fiat/src/fiat/system/internal/cxxdemangle.cc (code 2)
  ```
- NVHPC:
```
  FAILED: src/common.dp/CMakeFiles/cloudsc-common-dp-field_api.dir/module/yomphyder.F90.o src/common.dp/module-field_api/yomphyder.mod 
  /opt/nvhpc/Linux_x86_64/24.5/compilers/bin/nvfortran -I/home/runner/work/dwarf-p-cloudsc/dwarf-p-cloudsc/src/common/module -I/home/runner/work/dwarf-p-cloudsc/dwarf-p-cloudsc/src/common/include -I/home/runner/work/dwarf-p-cloudsc/dwarf-p-cloudsc/build/src/common.dp/module-field_api -I/home/runner/work/_temp/deps/hdf5/include -I/home/runner/work/_temp/install/field_api/include -I/home/runner/work/_temp/install/field_api/include/field_api_dp -I/home/runner/work/_temp/install/field_api/include/field_api -Mlarge_arrays -fast -O3 -fast -module src/common.dp/module-field_api -fPIC -mp -acc -c src/common.dp/CMakeFiles/cloudsc-common-dp-field_api.dir/module/yomphyder.F90-pp.f90 -o src/common.dp/CMakeFiles/cloudsc-common-dp-field_api.dir/module/yomphyder.F90.o
  NVFORTRAN-F-0004-Unable to open MODULE file oml_mod.mod (/home/runner/work/dwarf-p-cloudsc/dwarf-p-cloudsc/src/common/module/yomphyder.F90: 12)
  NVFORTRAN/x86-64 Linux 24.5-1: compilation aborted
  ```

For that reason I'm going to roll back the last commit and we can expand the testing in subsequent PRs.
For Atlas variant testing, I'm going to wait for @sbrdar's changes to become ready, at which point we should switch this on.

wertysas (approved):
Thank you for this makeover of the build system and general cleanup! I have tested building and executing the different Field API CPU and GPU variants and can confirm that that they work.

reuterbal:
Given that no objections have been raised, I will later today tag the current develop as a new release and then merge this PR.

mlange05 (approved):
Ok, first of all many thanks for a tremendous effort. This was really very much needed and was executed very neatly and elegantly! 🙏 🙌 🎉 

I've played with GPU, Python and C variants, and it all looks very nice (minor niggels sorted out offline). The only comment I have now is that the README really needs an update, as many options now take ON/OFF as a value. However, this can be left for another PR, so this here is GTG for me. :shipit: 

FWIW, I've also looked into Field-API and Atlas variants, and they all look good, except for Atlas + NVHPC-22.11. But I'm happy to leave that for a follow-on PR too.

reuterbal:
Thanks for the reviews! Any remarks @wdeconinck or shall we go ahead and do any fixes subsequently?

wdeconinck (approved):
This looks great @reuterbal , impressive work that will certainly increase the maintainability going forwards.
From the looks of this it looks OK. We can always fix up the atlas part later when the atlas-gpu-mutlifield contributions are finally ready.

PR #117, opened 2025-03-12
Add build-hpc workflow to run actions on HPC

reuterbal:


reuterbal:
OK, I believe this is ready now. The hpc-ci workflow side of things was pretty much straightforward thanks to the previous work on other repositories, and I have just fine-tuned the template for a bundle-based build process (so we have test coverage for that, too). 

I have piggy-backed a 24.5 arch for our HPC, for which an update seemed like a good idea anyway. 

I had to exclude a few tests which either sigfpe (NVHPC) or fail with segfaults on device (LUMI). These failures are reproducible interactively, therefore I figured their resolution should be addressed in a separate PR. On LUMI, I also got some weird OMPT initialization errors, so I had to build without OpenMP for now.

mlange05 (approved):
Looks fantastic and is very much appreciated! GTG from me. :shipit:

PR #118, opened 2025-04-04
Add Asynchronous and Blocked Field-API GPU variants

wertysas:
**[DO NOT MERGE]**

This PR adds two new GPU drivers and targets that use Field-API for device offload:
- ``dwarf-cloudsc-gpu-scc-field-blocked`` - double blocked driver with partial offload 
- ``dwarf-cloudsc-gpu-scc-field-async`` -  double blocked driver with asynchronous partial offload

The newly added variants introduce double blocked driver loops that offload the field in smaller chunks and therefore reduce the required allocation size on the device. This makes it possible to run much larger problem sizes compared to the other drivers.

### Field API dependency
**Note.** The new variants require Field API features that have not been merged into the main Field API repository yet but can be found on the fork https://github.com/wertysas/field_api. Therefore the bundle file has been updated to point to this branch on the fork for the time being.

https://github.com/ecmwf-ifs/dwarf-p-cloudsc/blob/c354cf6fcd4ff6a3a43881ee44e2678e1b92b29f/bundle.yml#L53-L55

### Building and runtime options

In order to build the blocked and async. Field API variants both CUDA (``--with-cuda=ON``) and Field API (``--with-field-api=ON``) must be enabled. The async functionality has been verified to work with NVHPC version 24.5. The variants can be used both with and without (``--without-mapped-fields``) OpenACC mapped pointers.

```
./cloudsc-bundle build --arch=arch/ecmwf/hpc2020/nvhpc/24.5 --with-cuda=ON --with-field-api=ON [--without-mapped-fields]
``` 

In order to get any benefits from using the asynchronous version, the fields should be allocated in pinned memory. This is controlled by setting the environment variable ``CLOUDSC_FIELD_API_PINNED=ON``. There are two more environment variables that can be used to control the chunk size and the number of queues used by blocked and async driver.
- ``CLOUDSC_BLOCK_SIZE`` - defaults to 512.
- ``CLOUDSC_ASYNC_NQUEUES`` - defaults to 3.

### Scaling and performance results

The graph below shows the performance numbers in Mflop/s as reported by the cloudsc timer for the variants with varying problem sizes. The measurements were made on an A100 one of ATOS GPU nodes. The code was compiled with the option ``--without--mapped-fields`` and the default values for the block size (512) and number of async queues (3) were used.

async-bench

Below is a screenshot showing the results of a run profiled with Nsight systems. Note that there is an overlap of compute and data transfers in both directions.

cloudsc-async-overlap


src/common/module/cloudsc_aux_type_mod.F90
```diff
@@ -12,7 +12,7 @@ MODULE CLOUDSC_AUX_TYPE_MOD
   USE YOECLDP,   ONLY : NCLV
 
   USE FIELD_MODULE, ONLY: FIELD_2RB, FIELD_3RB, FIELD_4RB, FIELD_2IM, FIELD_2LM, FIELD_3RB_PTR
-  USE FIELD_FACTORY_MODULE, ONLY: FIELD_NEW, FIELD_DELETE
```
mlange05, line 15: Why are we moving to unqualified imports here? (Same question for the other utilities.)

src/cloudsc_gpu/cloudsc_driver_gpu_scc_field_blocked_mod.F90
```diff
@@ -0,0 +1,296 @@
...
+
+      ! Local timer for each thread
+      TID = GET_THREAD_NUM()
+      CALL TIMER%THREAD_START(TID)
+
+!$acc parallel loop gang vector_length(NPROMA)
+    DO IBL=BLOCK_START,BLOCK_END ! just a way to loop over NGPBLKS
+        JKGLO=(IBL-1)*NPROMA+1
+        ICEND=MIN(NPROMA, NGPTOT-JKGLO+1)
+      ! DO JKGLO=1,NGPTOT,NPROMA
```
mlange05: Debug leftover?

src/cloudsc_gpu/cloudsc_driver_gpu_scc_field_async_mod.F90
```diff
@@ -0,0 +1,374 @@
...
+    CALL FLUX%F_PFHPSL%CREATE_DEVICE_DATA(BLK_BOUNDS=[1,NQUEUES*BLOCK_SIZE])
+    CALL FLUX%F_PFHPSN%CREATE_DEVICE_DATA(BLK_BOUNDS=[1,NQUEUES*BLOCK_SIZE])
+
+    CALL TENDENCY_LOC%F_T%CREATE_DEVICE_DATA(BLK_BOUNDS=[1,NQUEUES*BLOCK_SIZE])
+    CALL TENDENCY_LOC%F_Q%CREATE_DEVICE_DATA(BLK_BOUNDS=[1,NQUEUES*BLOCK_SIZE])
+    CALL TENDENCY_LOC%F_A%CREATE_DEVICE_DATA(BLK_BOUNDS=[1,NQUEUES*BLOCK_SIZE])
+    CALL TENDENCY_LOC%F_CLD%CREATE_DEVICE_DATA(BLK_BOUNDS=[1,NQUEUES*BLOCK_SIZE])
+
+
+    DO BLOCK_IDX=0, BLOCK_COUNT-1
```
mlange05: Uhhh, 0-offset counting, how un-Fortranic? 😉 I'm assuming this is to make the block-index calculations easier?

src/cloudsc_gpu/dwarf_cloudsc_gpu.F90
```diff
@@ -127,20 +145,45 @@ PROGRAM DWARF_CLOUDSC
...
+! Create a global memory state from serialized input data
 CALL GLOBAL_STATE%LOAD(NPROMA, NGPTOT, NGPTOTG, USE_PACKED=USE_PACKED)
+#elif defined(CLOUDSC_GPU_SCC_FIELD_BLOCKED)
+! Use Field API pinned memory capabilities for
+CALL GET_ENVIRONMENT_VARIABLE('CLOUDSC_FIELD_API_PINNED', PINNED_FIELD_MEMORY)
+INIT_PINNED_VALUE = TRIM(PINNED_FIELD_MEMORY) == 'ON' .OR. TRIM(PINNED_FIELD_MEMORY) == '1'
+! Create a global memory state from serialized input data
+CALL GLOBAL_STATE%LOAD(NPROMA, NGPTOT, NGPTOTG, USE_PACKED=.FALSE.)
+! Load block size from env.variable if present
+CALL GET_ENVIRONMENT_VARIABLE('CLOUDSC_BLOCK_SIZE', BLOCK_SIZE_CHARS, STATUS=ENV_STATUS)
```
mlange05: These options all need to be mentioned in the README, as they are specific to the distinct variants.

bundle.yml
```diff
@@ -51,8 +51,8 @@ projects :
             ECKIT_ENABLE_CUDA=OFF
 
     - field_api :
-        git     : https://github.com/ecmwf-ifs/field_api.git
-        version : v0.3.4
+        git     : https://github.com/wertysas/field_api.git
+        version : je-async-partial-offload-force
```
mlange05: Integration note: This obviously cannot be merged. As discussed offline, I think we'll need to set up a dedicated branch in FIELD API if we want to close the CI testing loop and run this on the `build-hpc` runners.

src/cloudsc_gpu/cloudsc_driver_gpu_scc_field_blocked_mod.F90
```diff
@@ -0,0 +1,296 @@
...
+          & PFSQRF(:,:,IBL),   PFSQSF (:,:,IBL),  PFCQRNG(:,:,IBL),  PFCQSNG(:,:,IBL),&
+          & PFSQLTUR(:,:,IBL), PFSQITUR (:,:,IBL), &
+          & PFPLSL(:,:,IBL),   PFPLSN(:,:,IBL),   PFHPSL(:,:,IBL),   PFHPSN(:,:,IBL),&
+          & YRECLDP=LOCAL_YRECLDP)
+
+      ENDDO
+!$acc end parallel loop
+!$acc end data
+
+      CALL TIMER%THREAD_END(TID)
```
mlange05: These interior timers now make no more sense, as they only time a single sub-block. Not sure if we want to keep them, or remove them, or even highjack the `TID` and record timings for each "macro block"? The current setting spews out misleading compute speed-ups though, which is not good.

src/cloudsc_gpu/dwarf_cloudsc_gpu.F90
```diff
@@ -73,20 +85,26 @@ PROGRAM DWARF_CLOUDSC
...
+CHARACTER(LEN=5)  :: BLOCK_SIZE_CHARS
 LOGICAL :: USE_PACKED
 
+INTEGER :: ENV_STATUS = 1              ! Status for reading env. variables
 INTEGER(KIND=JPIM) :: NUMOMP   = 1     ! Number of OpenMP threads for this run
 INTEGER(KIND=JPIM) :: NGPTOTG  = 16384 ! Number of grid points (as read from command line)
 INTEGER(KIND=JPIM) :: NPROMA   = 64   ! NPROMA blocking factor (currently active)
 INTEGER(KIND=JPIM) :: NGPTOT           ! Local number of grid points
+INTEGER(KIND=JPIM) :: NQUEUES = 3      ! Number of queues for async F-API variant
+INTEGER(KIND=JPIM) :: BLOCK_SIZE = 512 ! Default block size for blocked and async F-API variants
```
mlange05: For the "macro block" size this almost seems like a very small default value. Or is this the "number of NPROMA blocks"? In which case the variable name is somewhat misleading, no?

src/cloudsc_gpu/dwarf_cloudsc_gpu.F90
```diff
@@ -127,20 +145,45 @@ PROGRAM DWARF_CLOUDSC
...
+  READ( BLOCK_SIZE_CHARS, * ) BLOCK_SIZE
+END IF
+#elif defined(CLOUDSC_GPU_SCC_FIELD_ASYNC)
 ! Use Field API pinned memory capabilities for
 CALL GET_ENVIRONMENT_VARIABLE('CLOUDSC_FIELD_API_PINNED', PINNED_FIELD_MEMORY)
 INIT_PINNED_VALUE = TRIM(PINNED_FIELD_MEMORY) == 'ON' .OR. TRIM(PINNED_FIELD_MEMORY) == '1'
+! Create a global memory state from serialized input data
+CALL GLOBAL_STATE%LOAD(NPROMA, NGPTOT, NGPTOTG, USE_PACKED=.FALSE.)
+! Load number of queues from env. variable if present
+CALL GET_ENVIRONMENT_VARIABLE('CLOUDSC_ASYNC_NQUEUES', NQUEUES_CHARS, STATUS=ENV_STATUS)
```
mlange05, line 173: Does it make sense to expose this here? Does this work with an arbitrary value of, say, 8? From the code it looks like a classic triple-buffering logic. Would something like 2 then also work without race conditions?

mlange05 (changes requested):
Ok, first of all, a very big thank you and well done for this massive achievement. This is a very advanced optimisation and works impressively well! I've flagged a few inconsistencies that we should address first, but overall the approach looks very good and very informative and exciting. Once thing I would like is some README entries that explain the use cases and the additional flags for this (I know README needs some TLC, but let's get at least this one right).

I was wondering if two distinct driver variants were needed for this, but it seems the setup and buffer adjustments are sufficiently different between synced and asynchronous copies that his seems warranted. 

So, despite the various in-line comments, I think this is looking very good and once the comments have been addressed we should aim to bring this in and start CI testing this. The downstream implications of that we can discuss offline. Very well done and many thanks!

src/common/module/cloudsc_aux_type_mod.F90
joherics: There is a reason for this. After the refactor of the directory order and CMake updates in Field API (see PR: [Directory reorganisation #74](https://github.com/ecmwf-ifs/field_api/pull/74)) this led to compilation errors. I spoke offline with @awnawab about this and we agreed that a an easy work around for now is to use unqualified imports instead. But I believe this is a Field API issue.

We don't have this problem in develop because it is using an older Field API version [v0.3.4](https://github.com/ecmwf-ifs/field_api/releases/tag/v0.3.4) from before the directory reorganisation.

src/cloudsc_gpu/cloudsc_driver_gpu_scc_field_async_mod.F90
joherics: Yes, I think this makes the subsequent blocking variable calculation easier, but that is also because my C++ brain prefers 0-indexing. However, it makes sense to switch this to 1-based indexing and I will do that.

src/cloudsc_gpu/dwarf_cloudsc_gpu.F90
joherics: Yes, this would work with an arbitrary numbers of streams, since the ``D2H``-transfer, ``KERNEL CALL``, and ``H2D``-transfer that operate on the same data block always are queued in that order on the same stream. I would expect a speedup with 2 streams as well, since most of the speedup given by this variant is achieved by the overlap of ``D2H`` and ``H2D`` and the kernel execution time is so small compared to data transfers.

I also don't see any reason to use more than 3 streams in this case. By my understanding, the main reason for using more than 3 streams is when the compute time outweighs the data transfer time and multiple kernels can be executed concurrently on the device. However, I still think it makes sense to let the user play around with the number of streams and block sizes. In the end it's scheduler that controls the work anyway, and it possible that it could schedule the work in a more optimal way.

joherics: It actually seems that using only 2 streams compared 3 leads to slightly better performance. I have not had time to look at the profiles yet, but I think it shows that it is a good idea to make both the block size and the number of streams runtime configurable.

src/cloudsc_gpu/dwarf_cloudsc_gpu.F90
joherics: I can see that this is confusing, and yes, this is the size of the blocks in the final dimension of the field , which in this case is the number of ``NPROMA`` blocks. The Field-API force methods only support blocked transfers w.r.t. the final  dimension and the blocking indices passed to the Field-API force methods are with respect to the final dimension. Given this, I think that the name of the variable makes sense, but I would be happy to change it as well. Do you have any suggestions? :)

awnawab: I agree that the naming convention whilst logical, before we are "block"-ing the kernel, can be confusing. Possible alternatives could be: chunk, strip, tile. The F-API calls should retain `BLK_BOUNDS` as the argument though; a field is unaware of the "chunks" and just gets a range of "blocks" to operate on.

src/common/module/cloudsc_aux_type_mod.F90
awnawab: Yes I vaguely remember there was a problem, but I just tested again offline and we can safely make them qualified imports again.

src/common/module/cloudsc_field_state_mod.F90
```diff
@@ -24,7 +24,7 @@ MODULE CLOUDSC_FIELD_STATE_MOD
   USE VALIDATE_MOD, ONLY: VALIDATE
   USE CLOUDSC_MPI_MOD, ONLY: IRANK
   USE FIELD_MODULE, ONLY: FIELD_2RB, FIELD_3RB, FIELD_4RB, FIELD_2IM, FIELD_2LM, FIELD_3RB_PTR
-  USE FIELD_FACTORY_MODULE, ONLY: FIELD_NEW, FIELD_DELETE
+  USE FIELD_FACTORY_MODULE
```
awnawab: This one is unused though, and can just be deleted.

src/cloudsc_gpu/dwarf_cloudsc_gpu.F90
awnawab: The optimum number of streams and chunks/blocks will vary from problem to problem, so yes, making them runtime configurable (also in the Loki transformation) is a good idea.

src/cloudsc_gpu/cloudsc_driver_gpu_scc_field_blocked_mod.F90
awnawab: If I am not mistaken, placing the timer around the outer block loop should restore the same meaning to the timing output as all the other variants, in which case we should do that.

src/cloudsc_gpu/cloudsc_driver_gpu_scc_field_async_mod.F90
```diff
@@ -0,0 +1,374 @@
...
+      BLOCK_START=BLOCK_IDX*BLOCK_SIZE+1
+      BLOCK_END=MIN((BLOCK_IDX+1)*BLOCK_SIZE, NGPBLKS)
+      BLK_BOUNDS=[BLOCK_START, BLOCK_END]
+      QUEUE = MODULO(BLOCK_IDX,NQUEUES)+1
+      OFFSET = (QUEUE-1)*BLOCK_SIZE
+      CHUNK_SIZE = BLK_BOUNDS(2) - BLK_BOUNDS(1) + 1
```
awnawab: I am really happy with how small the control-flow diff is between this and the synchronous blocked/chunked version, which will be a boon for debugging when we start rolling this out. Great work 😄

src/cloudsc_gpu/cloudsc_driver_gpu_scc_field_async_mod.F90
```diff
@@ -0,0 +1,374 @@
...
+!$acc & PFSQLF,   PFSQIF ,  PFCQNNG,  PFCQLNG,&
+!$acc & PFSQRF,   PFSQSF ,  PFCQRNG,  PFCQSNG,&
+!$acc & PFSQLTUR, PFSQITUR , &
+!$acc & PFPLSL,   PFPLSN,   PFHPSL,   PFHPSN) &
+!$acc & async(QUEUE)
+      DO IBLLOC=1, CHUNK_SIZE
+        IBL= BLOCK_SIZE*BLOCK_IDX+IBLLOC
+        JKGLO=(IBL-1)*NPROMA+1
+        ICEND=MIN(NPROMA,NGPTOT-JKGLO+1)
+      ! DO JKGLO=1,NGPTOT,NPROMA ! just a way to loop over NGPBLKS
```
awnawab: Debug leftover.

bundle.yml
awnawab: As discussed offline, I have just created a `develop-ecwmf` branch that can be used to CI test these developments.

awnawab (changes requested):
Well done and many many thanks for this enormous and excellent piece of work 🙏 🙏 You've managed to implement very advanced and sophisticated functionality in really a very elegant way 👏 

I've left a few small comments, with the main one being a reconsideration of the naming scheme. I've also created a `develop-ecmwf` branch which you can file the accompanying F-API PR against, please do ping me once that is ready to review!

src/common/module/cloudsc_aux_type_mod.F90
awnawab: Scratch that, it seems this is an nvhpc 22.11 specific bug, but is fixed in 24.5. For the sake of backwards compatibility, we can keep the module import unqualified.

wertysas:
Thank you for the reviews @mlange05 @awnawab, I have addressed your comments now and also updated the FIELD API version in the bundle following the merge of  [[Partial Offload] Add Support for Blocked and Asynchronous Memory Transfers #94](https://github.com/ecmwf-ifs/field_api/pull/94) into ``ecmwf-ifs/field_api/develop-ecmwf``. Thus, I have converted this from a draft to a standard pull request and it is ready to be reviewed again.

Regarding the timer code, I have moved the "thread timer" to be outside the block loops (next to the global timer). This makes the output look like below.

```
     NUMPROC=1, NUMOMP=1, NGPTOTG=720000, NPROMA=128, NGPBLKS=5625
 Reference MFLOP count for 100 columns :  12.48232900
     NUMOMP    NGPTOT  #GP-cols     #BLKS    NPROMA tid# : Time(msec)  MFlops/s     col/s
          1    720000    720000         1       128    0 :       5807     15475    123978 @ rank#0:core#96
          1    720000    720000      5625       128   -1 :       5807     15475    123978 : TOTAL @ rank#0
    1 x   1    720000    720000      5625       128   -1 :       5807     15475    123978 : TOTAL
 Executing CLOUDSC-GPU, "SCC-FIELD-BLOCKED", BLOCKING_CHUNK_SIZE,          512
 PINNED FIELDS  F
 DEV_ALLOC_SIZE = 68719476736
             Variable Dim             MinValue             MaxValue            AbsMaxErr         AvgAbsErr/GP          MaxRelErr-%
                PLUDE 2D1  0.0000000000000E+00  0.1026720108982E-03  0.0000000000000E+00  0.0000000000000E+00  0.0000000000000E+00
             PCOVPTOT 2D3  0.0000000000000E+00  0.1000000000000E+01  0.3108624468950E-14  0.6721012635325E-15  0.2522121455387E-14
     PRAINFRAC_TOPRFZ 1D1  0.0000000000000E+00  0.0000000000000E+00  0.0000000000000E+00  0.0000000000000E+00  0.0000000000000E+00

      ...
```

I think this is the least confusing way of presenting the results (given that a thread timer measurement must be logged before calling ``TIMER%PRINT_PERFORMANCE``).

README.md
```diff
@@ -107,6 +107,24 @@ In order to do so, please create a pull request with your contribution and sign
   minor modifications (i.e. derived types/global paramters handling).
   Turned off by default, activate at the build stage with 
   `--cloudsc-fortran-pyiface=ON`.
+- **dwarf-cloudsc-gpu-scc-field-blocked**: A version that uses FIELD API for
+data offload and SCC the SCC loop layout, with a double blocked driver loop that
```
awnawab: typo, should just be "the SCC loop layout"

README.md
```diff
@@ -107,6 +107,24 @@ In order to do so, please create a pull request with your contribution and sign
...
+offloads the data in chunks and make it possible to run arbitrarily large problem sizes.
```
awnawab: reword: "and enables running arbitrarily large problem sizes."

README.md
```diff
@@ -107,6 +107,24 @@ In order to do so, please create a pull request with your contribution and sign
...
+Two environment variables can be used to configure this variant:
+  * ``CLOUDSC_BLOCKING_CHUNK_SIZE`` (default ``512``) - The chunk size (the number
+of of  ``NPROMA`` blocks) that will be processed in each iteration.
```
awnawab: typo: remove one "of"

README.md
```diff
@@ -107,6 +107,24 @@ In order to do so, please create a pull request with your contribution and sign
...
+  * ``CLOUDSC_FIELD_API_PINNED`` (default ``OFF``) - If turned ``ON`` fields will be allocated
+in page-locked memory.
+- **dwarf-cloudsc-gpu-scc-field-async**: A version that uses FIELD API for
+data offload and SCC the SCC loop layout, with a double blocked driver loop that
+offloads the data in chunks and make it possible to run arbitrarily large problem sizes. Three
```
awnawab: Same typo and rewording suggestion as above.

README.md
```diff
@@ -107,6 +107,24 @@ In order to do so, please create a pull request with your contribution and sign
...
+environment variables can be used to configure this variant:
+  * ``CLOUDSC_BLOCKING_CHUNK_SIZE`` (default ``512``) - The chunk size (the number
+of ``NPROMA`` blocks) that will be processed in each iteration.
+  * ``CLOUDSC_ASYNC_NQUEUES`` (default ``3``) - The number of queues used by the async variant.
```
awnawab: reword: number of queues (i.e. CUDA streams)

README.md
```diff
@@ -107,6 +107,24 @@ In order to do so, please create a pull request with your contribution and sign
...
+  * ``CLOUDSC_FIELD_API_PINNED`` (default ``OFF``) - If CUDA ie enabled this variable **should be set to ``ON``**.
```
awnawab: typo: If CUDA *is* enabled

awnawab (approved):
Thanks for addressing the suggested changes. This looks fantastic now 👌 I've just left a few comments on the README, but otherwise I'm very happy with it so already approving 🙏

wertysas:
> Thanks for addressing the suggested changes. This looks fantastic now 👌 I've just left a few comments on the README, but otherwise I'm very happy with it so already approving 🙏

Thank you for the suggestions and for pointing out the typos! I have corrected this and updated the README.

mlange05 (approved):
This looks fantastic now and is ready to be merged. Again, a great piece of work and very advanced optimisation, elegantly implemented very well done. GTG for me.

PR #119, opened 2025-05-06
Bugfix: Remove zeroing of fields inside driver for FIELD API CPU variants

wertysas:
This is a simple bug fix PR that fixes an error in the output of the FIELD API CPU variant.  This PR removes zeroing of ``PAUX%PCOVTOT`` and ``TENDENCY_LOC%CLD`` inside the driver loop as it led to errors in the cloudsc output. The fields are already initialised to zero in the respective state types and need not be initialised inside the driver.


reuterbal (approved):

reuterbal:
Uncontroversial, looks good to me

mlange05 (approved):
Agreed, uncontroversial. 👍 :shipit:

PR #121, opened 2025-06-14
New IO capabilities via FieldAPI IO 

piotrows:
This PR adds HDF5 IO feature to a FieldAPI variant of CLOUDSC. The changes proposed are:

1.  Extra routine _TEST_FIELD_HDF5_ placed next to the **field's** driver version of the Fortran feature, performing FieldAPI HDF5 calls to write and subsequently read CLOUDSC input data. The routine queries for host pointers of the FieldAPI variables, then stores all in the HDF5 file (or files, in case of MPI variant), and ultimately reads them again to verify if the write/read was performed correctly.
2.  Fiat is added as a FieldAPI dependency, since FieldAPI IO module depends on MPI for generality. Fiat-specific files duplicated in CLOUDSC dwarf are compiled only if Fiat is FieldAPI is requested to be disabled.


reuterbal:
The HDF5 functionality has now been added to field_api main. Could this also be merged into develop_ecmwf, as that's the branch we use in cloudsc? (@awnawab)

piotrows:
> The HDF5 functionality has now been added to field_api main. Could this also be merged into develop_ecmwf, as that's the branch we use in cloudsc? (@awnawab)

Yes please! I need it for further developments.

awnawab:
> The HDF5 functionality has now been added to field_api main. Could this also be merged into develop_ecmwf, as that's the branch we use in cloudsc? (@awnawab)

Sure, I'll rebase ecmwf-develop over main.

awnawab:
F-API develop-ecmwf now rebased over main.

reuterbal:
@piotrows Can you update the field_api branch in the bundle.yml and resolve the conflicts? Then this should be ready for a look I believe?

piotrows:
> @piotrows Can you update the field_api branch in the bundle.yml and resolve the conflicts? Then this should be ready for a look I believe?

Done, let's have a look together. I've made this development a bit more optional.

src/cloudsc_fortran/cloudsc_driver_field_mod.F90
```diff
@@ -115,6 +120,207 @@ SUBROUTINE CLOUDSC_DRIVER_FIELD( &
...
+    
+    USE FIELD_HDF5_MODULE, ONLY: WRITE_HDF5_PERRANK_DATA, READ_HDF5_PERRANK_DATA
+    INTEGER(KIND=JPIM)        ,INTENT(IN) :: NUMOMP, NPROMA, NLEV, NGPTOT, NGPTOTG
+    INTEGER(KIND=JPIM)        ,INTENT(IN) :: KFLDX 
+    REAL(KIND=JPRB)           ,INTENT(IN) :: PTSPHY       ! PHYSICS TIMESTEP
+    TYPE(CLOUDSC_AUX_TYPE)    ,INTENT(IN) :: PAUX
+    TYPE(CLOUDSC_STATE_TYPE)  ,INTENT(IN) :: TENDENCY_TMP
+    TYPE(CLOUDSC_STATE_TYPE)  ,INTENT(IN) :: TENDENCY_LOC
+
+    REAL(KIND=JPRB), POINTER    :: PAUX_PT(:,:,:)
```
mlange05: I don't think the pointer declarations are needed here, or am I missing something?

src/cloudsc_fortran/cloudsc_driver_field_mod.F90
```diff
@@ -50,6 +50,11 @@ SUBROUTINE CLOUDSC_DRIVER_FIELD( &
       write(0,1003) NUMPROC,NUMOMP,NGPTOTG,NPROMA,NGPBLKS
     end if
 
+#ifdef FIELD_API_IO
```
mlange05, line 52: To me, the option to write the input state to file would be a run-time option, and not a compile-time option. How about we rename `TEST_FIELD_HDF5` to `WRITE_STATE_HDF5` and make it a property of the `CLOUDSC_FIELD_STATE` instead of invoking it from the kernel? This way we could toggle this via an environment flag in `src/cloudsc_fortran/dwarf_cloudsc.F90` right after loading the input state from the classic input files? 

Similar we could then write the output by implementing the same boilerplate on `CLOUDSC_FLUX_TYPE` instead of adding the I/O to the kernel driver.

bundle.yml
```diff
@@ -100,6 +105,14 @@ options :
         help : Enable Field API variants [ON|OFF]
         cmake : BUILD_field_api={{value}}
 
+    - with-field-api-io :
+        help : Enable Field API IO (requires MPI) [ON|OFF]
+        cmake : >
+            FIELD_API_ENABLE_IO={{value}}
+            FIELD_API_ENABLE_FIAT={{value}}
```
mlange05: Should we not always enable FIAT now that it is part of the bundle? Also, could we sensibly always enabled Field-API-IO, or do we need that toggle for installs without hdf5?

src/common/CMakeLists.txt
```diff
@@ -58,13 +61,17 @@ macro( cloudsc_add_common_lib )
...
             $<${HAVE_HDF5}:module/hdf5_file_mod.F90>
             $<${HAVE_MPI}:module/cloudsc_mpif.F90>
         PUBLIC_LIBS
             $<${HAVE_HDF5}:hdf5::hdf5_fortran>
             $<${HAVE_SERIALBOX}:Serialbox::Serialbox_Fortran>
             $<${HAVE_MPI}:MPI::MPI_Fortran>
             ${COMMON_LIB_LIBS}
+            $<${HAVE_FIELD_API}:fiat>
+            $<${HAVE_FIELD_API}:field_api_${prec}>
+#            $<$<NOT:$<BOOL:${HAVE_FIELD_API}>>:parkind_${prec}>
```
mlange05: Debug leftover?

mlange05 (commented):
First of all, apologies for the delayed review comments, and second of all many thanks for taking this on. I think this is already very useful and I've run a basic example locally and things do indeed work as I'd expect them to. So, many thanks for this and much appreciated.

Next up, I've left a few comments mostly about structure and where to call what. In short, I think it would be better if we would add the file-write capabilities as utilities of the field-based state and flux objects, and call these outside of the driver layer, triggered by env variables to allow run-time switching instead of compile-time switching.

I've also noted two things:
* Nothing in the tests actually tests the correct execution of the output file, so silent failure would not be caught. Maybe a simple check for the existence of the dumped file(s) could be added to one of the ctests?
* Executing the same run twice only works with a data set of the same size, however changing the working set size on a re-run fails. I'm assuming we want to replace existing output files on the re-run instead of writing into the existing file?

bundle.yml
piotrows: @reuterbal was strongly opposing permanent dependence on FIAT. 
Since the availability of HDF5 for non-standard compilers is not obvious, perhaps we could stick to an optional HDF5...

piotrows:
> * Nothing in the tests actually tests the correct execution of the output file, so silent failure would not be caught. Maybe a simple check for the existence of the dumped file(s) could be added to one of the ctests?

Actually, there is a check. If you look closely, there is first a store to the output file and then an immediate read. If the store is corrupt, the read shall be also corrupt and the usual norms will fail.

piotrows:
> * Executing the same run twice only works with a data set of the same size, however changing the working set size on a re-run fails. I'm assuming we want to replace existing output files on the re-run instead of writing into the existing file?

This is a design question ... the current solution prevents the user from overwriting his previously stored data and was chosen deliberately. We can extend the API to support both ways, or we can change the default behavior.

src/cloudsc_fortran/cloudsc_driver_field_mod.F90
piotrows: > How about we rename `TEST_FIELD_HDF5` to `WRITE_STATE_HDF5`

Could be done but what we are doing at the moment is testing (write+read, not a write alone).

src/cloudsc_fortran/cloudsc_driver_field_mod.F90
piotrows: They are needed for the FA IO API so the shape of the variable becomes explicit.

src/cloudsc_fortran/cloudsc_driver_field_mod.F90
piotrows: > make it a property of the `CLOUDSC_FIELD_STATE` instead of invoking it from the kernel?

Well, this is an option, but then it looses the virtue of a top-level, clear example that teaches the FieldAPI IO. So it really depends if we want expand CLOUDSC capabilities or we would rather prefer a FA IO tutorial, or perhaps both.

piotrows:
@mlange05  @reuterbal 
Quick summary of where we are, feel free to further comment:
1. BR suggested (offline) that I should get rid of the quest for replacing Fiat files with Fiat lib, because it makes things complicated and the Atos CI runs fail due to linking errors. I am not fully convinced (perhaps we could link Fiat static to satisfy the GPU compiler needs), but certainly we have a decision to make.
2. I have a feeling that HDF5 should remain optional for CLOUDSC at compile time, but run-time is fine to me as well (HDF5 2.x (in devel) shall get rid of autotools and be easier to install, I believe)
3. Explicit example of FA IO vs. deeper integration with CLOUDSC aiming at expanding CLOUDSC capabilities, or both.
4. Do we need another evaluation of the HDF5 store correctness.

src/cloudsc_fortran/cloudsc_driver_field_mod.F90
mlange05: Are they truly needed there though? From my reading of https://github.com/ecmwf-ifs/field_api/blob/main/src/io/field_RANKSUFF_hdf5_module.fypp#L50 it uses `SELF%PTR` throughout, so there's no need to pass an externally declared pointer in. I also think this makes it look much more cumbersome than it needs to be, as the field object should have all the information locally to just write its content.

src/cloudsc_fortran/cloudsc_driver_field_mod.F90
mlange05: > Well, this is an option, but then it looses the virtue of a top-level, clear example that teaches the FieldAPI IO. So it really depends if we want expand CLOUDSC capabilities or we would rather prefer a FA IO tutorial, or perhaps both.

I disagree on this one - having a pure procedural test routine does not make a good "tutorial". Instead, the derived types `CLOUDSC_FIELD_STATE` and `CLOUDSC_FLUX_STATE` are intended as examples of group types used in the EC-physics context. Adding independent read/write routines to those types is the main goal here, so why not illustrate this here in exactly this form?

bundle.yml
mlange05: Ok, that makes sense. In that case please leave this as is.

mlange05:
> @mlange05 @reuterbal Quick summary of where we are, feel free to further comment:
> 
> 1. BR suggested (offline) that I should get rid of the quest for replacing Fiat files with Fiat lib, because it makes things complicated and the Atos CI runs fail due to linking errors. I am not fully convinced (perhaps we could link Fiat static to satisfy the GPU compiler needs), but certainly we have a decision to make.
> 2. I have a feeling that HDF5 should remain optional for CLOUDSC at compile time, but run-time is fine to me as well (HDF5 2.x (in devel) shall get rid of autotools and be easier to install, I believe)
> 3. Explicit example of FA IO vs. deeper integration with CLOUDSC aiming at expanding CLOUDSC capabilities, or both.
> 4. Do we need another evaluation of the HDF5 store correctness.

1. Agreed. Let's keep this a PR for one feature and only add what's needed for that.
2. Agreed, let's keep the compile-time option. I'd still prefer the actual use of it to mimic the binary I/O and use run-time flags for that.
3. Deep integration was always the aim, as we don't need a second unit test (F-API unit tests this already). Ideally it would behave like the binary I/O and either optionally or always replace this field F-API enabled drivers.
4. I don't think the in-situ test is the right way. I'd prefer to have a ctest that checks if the file was created to test the writes, and maybe let F-API-based drivers read input/reference from HDF5 to test reads (either optionally or always)?

src/cloudsc_fortran/cloudsc_driver_field_mod.F90
```diff
@@ -115,6 +120,207 @@ SUBROUTINE CLOUDSC_DRIVER_FIELD( &
       CALL TIMER%PRINT_PERFORMANCE(NPROMA, NGPBLKS, NGPTOT)
     
   END SUBROUTINE CLOUDSC_DRIVER_FIELD
+#ifdef FIELD_API_IO 
+  SUBROUTINE TEST_FIELD_HDF5( &
+     & NUMOMP, NPROMA, NLEV, NGPTOT, NGPTOTG, KFLDX, PTSPHY, PAUX, &
+     & TENDENCY_TMP, TENDENCY_LOC)
+    ! Driver routine that invokes the optimized CLAW-based CLOUDSC GPU kernel
```
reuterbal: This seems not accurate
```suggestion
```

src/cloudsc_fortran/cloudsc_driver_field_mod.F90
reuterbal: I'm guessing this is required to have shape information embedded into the subroutine interface that allows Fortran to pick the correct variant for the field?

If the pointer was removed, something like
```
CALL WRITE_HDF5_PERRANK_DATA(PAUX%F_PT, PAUX_PVERVEL, "cloudsc_data", "PT",           HDFEXISTS=.TRUE.)
```
would then have to have either a dimension keyword
```
CALL WRITE_HDF5_PERRANK_DATA(PAUX%F_PT, 3, "cloudsc_data", "PT",           HDFEXISTS=.TRUE.)
```
or specify the actual implementation in the routine name, hypothetically paraphrasing:
```
CALL WRITE_HDF5_PERRANK_DATA_3RB(PAUX%F_PT, "cloudsc_data", "PT",           HDFEXISTS=.TRUE.)
```

I might be misinterpreting this, but if the above is the case, then pointers with a certain dimension are an unintuitive way of specifying this information. @awnawab might have a suggestion here?

src/common/CMakeLists.txt
```diff
@@ -9,25 +9,28 @@
...
     module/cloudsc_global_state_mod.F90
     module/cloudsc_mpi_mod.F90
 )
 
+set(CLOUDSC_FIAT_SOURCES
+    module/parkind1.F90
+    module/ec_pmon_mod.F90
+    module/oml_mod.F90
+    module/abor1.F90
+)
```
reuterbal: My suggestion would be to roll back this change and restore these files in the list above.

```suggestion
```

src/common/CMakeLists.txt
```diff
@@ -58,13 +61,17 @@ macro( cloudsc_add_common_lib )
         SOURCES
             ${CLOUDSC_COMMON_SOURCES}
             ${COMMON_LIB_SOURCES}
+            $<$<NOT:$<BOOL:${HAVE_FIAT}>>:${CLOUDSC_FIAT_SOURCES}>
```
reuterbal: This can then also be removed
```suggestion
```

src/common/CMakeLists.txt
reuterbal: And this is redundant in any case because the field_api variant that needs to be linked against field_api does specify this already:
https://github.com/ecmwf-ifs/dwarf-p-cloudsc/blob/4f1f7e5fa6a4b7f76b7c306379d2e3d991a5ae6d/src/common/CMakeLists.txt#L131-L135

bundle.yml
```diff
@@ -54,14 +54,19 @@ projects :
...
+        version : 1.5.1 
         require : ecbuild
         optional: true
+            
+    - field_api :
+        git     : https://github.com/ecmwf-ifs/field_api.git
+        version : develop-ecmwf 
+        require : ecbuild fiat
         cmake   : >
             UTIL_MODULE_PATH=${CMAKE_SOURCE_DIR}/cloudsc-dwarf/src/common/module
```
reuterbal: With fiat in the bundle, field_api shouldn't require the util_module_path anymore
```suggestion
```

bundle.yml
```diff
@@ -54,14 +54,19 @@ projects :
             ECKIT_ENABLE_ECKIT_SQL=OFF
             ECKIT_ENABLE_WARNINGS=OFF
 
-    - field_api :
-        git     : https://github.com/ecmwf-ifs/field_api
-        version : develop-ecmwf
+    - fiat      :
```
reuterbal, line 58: As discussed, fiat should be optional to keep the dwarf itself as minimal as possible, which has proven useful in the past. Thus, add a `BUILD_fiat=OFF` in l. 13 to disable fiat by default.

bundle.yml
```diff
@@ -100,6 +105,14 @@ options :
         help : Enable Field API variants [ON|OFF]
         cmake : BUILD_field_api={{value}}
```
reuterbal, line 111: Here, we should then enable fiat alongside field_api

```suggestion
        cmake : BUILD_fiat={{value}} BUILD_field_api={{value}}
```

CMakeLists.txt
```diff
@@ -133,6 +133,14 @@ ecbuild_add_option( FEATURE FIELD_API
                     DEFAULT ON
                     REQUIRED_PACKAGES "field_api" )
 
+ecbuild_add_option( FEATURE FIELD_API_ENABLE_IO
+		    DESCRIPTION "Use field_api IO to store field_api objects to disk."
+                    CONDITION HAVE_FIELD_API
+		    DEFAULT OFF )
```
reuterbal: Mixture of tabs and spaces
```suggestion
                    DESCRIPTION "Use field_api IO to store field_api objects to disk."
                    CONDITION HAVE_FIELD_API
                    DEFAULT OFF )
```

src/cloudsc_fortran/cloudsc_driver_field_mod.F90
reuterbal: How about splitting the two into a write and read routine and calling them in this order from here?
That would work as a demonstration of how to use either mode individually. 

For a full-on switch to HDF5 as inputs, we would need to replicate the inputs in a third format commited to the repository. I think that is a good idea but could also be tackled in a follow-on PR maybe?
The input reading is still very much set-up in a Serialbox first, otherwise use the legacy hdf5 read method. We could first remove that and full-on switch to HDF5, then add the field_api hdf5 reader.

reuterbal (changes requested):
Many thanks, overall this looks very nice. I agree with the API and integration points that @mlange05 raised and have also pointed out how to fix the fiat/field_api integration.

awnawab:
F-API develop-ecmwf rebased over main and now also contains https://github.com/ecmwf-ifs/field_api/pull/108.

bundle.yml
```diff
@@ -98,7 +107,15 @@ options :
...
-        cmake : BUILD_field_api={{value}}
+        cmake : BUILD_fiat={{value}}  BUILD_field_api={{value}}
+
+    - with-field-api-io :
+        help : Enable Field API IO (requires MPI) [ON|OFF]
+        cmake : >
+            FIELD_API_ENABLE_IO={{value}}
+            FIELD_API_ENABLE_FIAT={{value}}
+            ENABLE_FIELD_API_ENABLE_IO={{value}}
+            ENABLE_MPI=ON
```
reuterbal: Is there really a need to forcefully enable MPI?

piotrows: > Is there really a need to forcefully enable MPI?

The way it is coded in FA at the moment, yes. I am working on it on a FA side.

piotrows: > Is there really a need to forcefully enable MPI?

@reuterbal @awnawab  I've made MPI/MPL fully optional, what do you think about the current state?

src/cloudsc_cuda/CMakeLists.txt
```diff
@@ -31,6 +31,7 @@ macro( cloudsc_add_cuda_exe TARGET DRIVER KERNEL )
             $<${HAVE_HDF5}:hdf5::hdf5>
             $<${HAVE_SERIALBOX}:Serialbox::Serialbox_C>
             $<${HAVE_OMP}:OpenMP::OpenMP_C>
+            $<${HAVE_MPI}:MPI::MPI_C>
```
reuterbal, line 34: Why is this needed? The MPI use is only in the Fortran layer.

src/cloudsc_cuda/CMakeLists.txt
```diff
@@ -40,6 +41,8 @@ macro( cloudsc_add_cuda_exe TARGET DRIVER KERNEL )
     target_compile_options( ${TARGET}-${prec}-lib
         PRIVATE $<$<COMPILE_LANGUAGE:CUDA>:SHELL:${CLOUDSC_CUDA_OPT_FLAGS} ${CLOUDSC_CUDA_FLAGS}>
     )
+    target_link_libraries(${TARGET}-${prec}-lib PRIVATE $<${HAVE_MPI}:MPI::MPI_C>)
```
reuterbal: This should be redundant to the addition above (if that is even needed)

src/cloudsc_fortran/cloudsc_driver_field_mod.F90
```diff
@@ -115,6 +121,119 @@ SUBROUTINE CLOUDSC_DRIVER_FIELD( &
       CALL TIMER%PRINT_PERFORMANCE(NPROMA, NGPBLKS, NGPTOT)
     
   END SUBROUTINE CLOUDSC_DRIVER_FIELD
+#ifdef FIELD_API_IO 
+  SUBROUTINE WRITE_FIELD_HDF5( &
+     & NUMOMP, NPROMA, NLEV, NGPTOT, NGPTOTG, KFLDX, PTSPHY, PAUX, &
+     & TENDENCY_TMP, TENDENCY_LOC)
+    ! Driver routine that invokes the optimized CLAW-based CLOUDSC GPU kernel
```
reuterbal: I still think this comment is not correct here ;-)
```suggestion
```

src/common/CMakeLists.txt
```diff
@@ -56,15 +60,20 @@ macro( cloudsc_add_common_lib )
             $<$<STREQUAL:"${prec}","sp">:SINGLE>
             ${COMMON_LIB_DEFINITIONS}
         SOURCES
+            ${CLOUDSC_FIAT_SOURCES}
             ${CLOUDSC_COMMON_SOURCES}
             ${COMMON_LIB_SOURCES}
+#            $<$<NOT:$<BOOL:${HAVE_FIAT}>>:${CLOUDSC_FIAT_SOURCES}>
```
reuterbal: Please remove
```suggestion
```

src/common/CMakeLists.txt
```diff
@@ -56,15 +60,20 @@ macro( cloudsc_add_common_lib )
...
             $<${HAVE_HDF5}:module/hdf5_file_mod.F90>
             $<${HAVE_MPI}:module/cloudsc_mpif.F90>
         PUBLIC_LIBS
             $<${HAVE_HDF5}:hdf5::hdf5_fortran>
             $<${HAVE_SERIALBOX}:Serialbox::Serialbox_Fortran>
             $<${HAVE_MPI}:MPI::MPI_Fortran>
             ${COMMON_LIB_LIBS}
+#            $<${HAVE_FIELD_API}:fiat>
+#            $<${HAVE_FIELD_API}:field_api_${prec}>
+#            $<$<NOT:$<BOOL:${HAVE_FIELD_API}>>:parkind_${prec}>
```
reuterbal: Please remove
```suggestion
```

bundle.yml
```diff
@@ -54,20 +55,29 @@ projects :
...
         cmake   : >
             FCKIT_ENABLE_TESTS=OFF
+            FCKIT_ENABLE_FCKIT_VENV=ON
+
+    - field_api :
+        git     : https://github.com/ecmwf-ifs/field_api.git
+        version : napz-IO-mpi-independent-option 
+        require : ecbuild
+        cmake   : >
+            UTIL_MODULE_PATH=${CMAKE_SOURCE_DIR}/cloudsc-dwarf/src/common/module
```
reuterbal, line 79: I don't think we need this anymore if we build fiat
```suggestion
```

CMakeLists.txt
```diff
@@ -45,7 +45,7 @@ include(features/OMP)
 ecbuild_add_option( FEATURE MPI
                     DESCRIPTION "Support for task-level parallelism using MPI"
                     DEFAULT OFF
-                    REQUIRED_PACKAGES "MPI COMPONENTS Fortran" )
+                    REQUIRED_PACKAGES "MPI COMPONENTS Fortran C CXX" )
```
reuterbal, line 48: Still curious what makes this necessary?

CMakeLists.txt
```diff
@@ -133,6 +133,23 @@ ecbuild_add_option( FEATURE FIELD_API
...
+                    DESCRIPTION "Use field_api IO to store field_api objects to disk."
+                    CONDITION HAVE_FIELD_API
+                    DEFAULT OFF )
+
+if( HAVE_FIELD_API_ENABLE_IO )
+   list(APPEND CLOUDSC_DEFINITIONS FIELD_API_IO)
+   if( HAVE_FIELD_API_ENABLE_IO_MPL )
+     list(APPEND CLOUDSC_DEFINITIONS FIELD_API_IO_MPL)
+   elseif( HAVE_FIELD_API_ENABLE_IO_MPI )
+     find_package(MPI REQUIRED )
```
reuterbal: I think it makes more sense to check here that the MPI feature is enabled. The bundle implicitly enforces this but a CMake-build without bundle would allow a halfway-house otherwise

```suggestion
     if( NOT HAVE_MPI )
         ecbuild_error("Cannot enable FIELD_API_ENABLE_IO_MPI when feature MPI is disabled.")
     endif()
```

CMakeLists.txt
```diff
@@ -133,6 +133,23 @@ ecbuild_add_option( FEATURE FIELD_API
...
+     list(APPEND CLOUDSC_DEFINITIONS FIELD_API_IO_MPI)
+   elseif( HAVE_FIELD_API_ENABLE_IO_SERIAL )
```
reuterbal: Should this be the default if none of the MPI/MPL options are enabled?
Again, relevant for the non-bundle CMake build in particular.

README.md
```diff
@@ -35,7 +35,10 @@ In order to do so, please create a pull request with your contribution and sign
   also is slightly cleaner than the original version.
 - **dwarf-cloudsc-fortran-field**: A fortran version of CLOUDSC that uses Field API
   for the data structures. The intent of this version is to show how
-  Field API is used in newer versions of the IFS.
+  Field API is used in newer versions of the IFS. Optionally, this version also tests 
+  Field  API IO.  To test IO write/read feature,  add --with-field-api-io at the build stage.
+  Note that this enables MPI by default. The IO feature is embedded in the 
+  cloudsc-fortran Field API test.
```
reuterbal, line 41: This probably needs a small update later.

CMakeLists.txt
```diff
@@ -133,6 +133,23 @@ ecbuild_add_option( FEATURE FIELD_API
...
 
+ecbuild_add_option( FEATURE FIELD_API_ENABLE_IO
+                    DESCRIPTION "Use field_api IO to store field_api objects to disk."
+                    CONDITION HAVE_FIELD_API
+                    DEFAULT OFF )
+
+if( HAVE_FIELD_API_ENABLE_IO )
+   list(APPEND CLOUDSC_DEFINITIONS FIELD_API_IO)
+   if( HAVE_FIELD_API_ENABLE_IO_MPL )
+     list(APPEND CLOUDSC_DEFINITIONS FIELD_API_IO_MPL)
```
reuterbal: I don't see the FIELD_API_IO_MPL, FIELD_API_IO_MPI, or FIELD_API_IO_SERIAL definitions being used anywhere in the code. Are they required here?

reuterbal (changes requested):

bundle.yml
reuterbal: That's great, many thanks! Looks very nice on the CLOUDSC side now, but I haven't look at Field-API. I'll let @awnawab judge the F-API angle ;-)

src/cloudsc_cuda/CMakeLists.txt
piotrows: This is a tricky one. If HDF5 is compiled with a parallel option, then hdf5.h demands mpi.h.

reuterbal: Ah, I see, thanks for clarifying. Fine with me then, but I think adding this once should then be sufficient?

bundle.yml
awnawab: I think we can simplify it a little in F-API. My comments that follow are based on what I saw in napz-IO-mpi-independent-option, please disregard if you had plans to change that.

Right now the FIELD_API IO feature requires fiat and for fiat to be built with MPI. For the moment, let's keep these two assumptions, and if in the future they become limitations we can address them then.

The only MPI related information we need in the IO functionality is the MPI communicator. The current rank and total number of ranks can be queried from the communicator. So I would propose we simply make the MPI communicator an optional argument to any IO routine that needs it. If this argument is not present, then we can assume serial IO. We can print a message either way to NULOUT so the user isn't surprised.

This allows us to remove all the IO related options: `IO_SERIAL, IO_MPI, IO_MPL`. FIAT links to MPI publicly, so linking to FIAT means we can import the MPI runtime library too in F-API. Since that is currently a requirement for the IO feature, we don't even need to ifdef guard it.

Apologies if I've missed something, and we can also chat about it offline if you'd like 😄

awnawab: So with the above, F-API IO still requires an MPI enabled FIAT, but user code like CLOUDSC can still use the IO functionality without requiring FIAT or MPL. That's really what I meant the other day, sorry if I wasn't clear. We shouldn't need to enforce MPL on user code in order to use F-API IO.

piotrows: > Right now the FIELD_API IO feature requires fiat and for fiat to be built with MPI.

But the whole point of this recent development is to remove Fiat dependence and MPI dependence completely from the HDF5 IO functionality, following BR question if the MPI dependence is really needed.

piotrows: > The only MPI related information we need in the IO functionality is the MPI communicator.

We definitely need a true rank number to compose/decompose field properly using prospective truly parallel HDF5 write.

piotrows: > The current rank and total number of ranks can be queried from the communicator.

Do we have a guarantee that the MPI and MPL rank numbers always match?

piotrows: > we simply make the MPI communicator an optional argument to any IO routine that needs it.

If you make it an argument, you need to always specify the TYPE(MPI_Comm), so you always need to link to MPI. To me saying that communicator is integer is not future-proof. Am I missing something here?

piotrows: > FIAT links to MPI publicly, so linking to FIAT means we can import the MPI runtime library too in F-API.

I understand that it doesn't have to link to MPI.

piotrows: > This allows us to remove all the IO related options: `IO_SERIAL, IO_MPI, IO_MPL`

Well, but this is a bit of code in cmake and a single routine vs. extending argument list, which should be kept minimal. The current trend seems to be to hide the MPI layer rather than repetitively introduce MPI_COMM_WORLD in each HDF5 store call.

piotrows: > So with the above, F-API IO still requires an MPI enabled FIAT,

This becomes confusing, because the proper code is in a different branch right now... perhaps I should force push the new content here, but some cmake clean-up the BR pointed out needs to go first. The idea is that FA IO should and could  not depend on Fiat and MPI at all.

awnawab: Here is how it could look:
```
CMakeLists.txt
---------------

ecbuild_add_option( FEATURE MPI
                    DESCRIPTION "Support for MPI distributed parallelism"
                    REQUIRED_PACKAGES "MPI COMPONENTS Fortran" )
if( HAVE_MPI)
  list(APPEND FIELD_API_DEFINITIONS WITH_MPI)
endif()
## add MPI to libraries privately linked to F-API

hdf5_common_module
-----------------------

SUBROUTINE NAME_HDF5_STORAGE(FILENAME,FNAME_WITH_RANK &
#ifdef WITH_MPI
& ,MPI_COMM &
#endif
& )

#ifdef WITH_MPI
USE MPI, ONLY: ...
#endif
IMPLICIT NONE
#ifdef WITH_MPI
TYPE(MPI_COMM), INTENT(IN), OPTIONAL :: MPI_COMM
#endif

! query rank from communicator, which by default starts from 0
! MPL just increments this by 1, so we do the same here
....

MY_RANK = MY_RANK + 1

END SUBROUTINE
```

awnawab: We can also do it without any extra arguments. We can always import `USE MPI, ONLY : MPI_COMM_WORLD`, like you have it for the `IO_MPI` option, regardless of whether the user code is using MPL. It just means we increment the rank by 1 by default, for Fortran based programs 1 is probably the right choice to start the MPI ranks from (MPL also queries the rank from the communicator and adds 1).

Having 4 IO related options seems like overkill. An `IO` option coupled with an `MPI` option should be enough., and allows us to be independent of FIAT.

reuterbal: I think you misunderstood me. Having fiat as a dependency when we're building against field api is perfectly valid. I just don't like having fiat _always_ as a dependency on non-fapi targets.

piotrows: > It just means we increment the rank by 1 by default,

If it is not set in stone in MPL, then I would see it as unsafe to encode such an implicit assumption into non-MPL code.

piotrows: > Having 4 IO related options seems like overkill

I think there are 3: SERIAL, MPI, MPL. There need to be at least two specified at the compile time (serial and parallel), the third one merely allows to be independent of FIAT.

piotrows: > An `IO` option coupled with an `MPI` option should be enough., and allows us to be independent of FIAT.

Again, from a domain scientist point of view, there should be an option of storing the data in an organized manner that translates in an unique way to a computational mesh, so the h5 file can be e.g. viewed in an external viewer or auto-processed in Python for postprocessing. At least until we move on to Atlas.

piotrows: > I think you misunderstood me. Having fiat as a dependency when we're building against field api is perfectly valid. I just don't like having fiat _always_ as a dependency on non-fapi targets.

My main point was to provide an MPI-independent setup, actually. You had a good point that this should be relaxed and indeed I see that for the sake of testing new compilers it is very handy to don't rely on the accompanying MPI.

CMakeLists.txt
piotrows: Actually, at the moment it is activated in FieldAPI itself:
elseif(HAVE_IO_MPI)
ecbuild_enable_mpi( COMPONENTS FORTRAN REQUIRED)

so this section of cloudsc can be deleted indeed.
However, if we link to parallel version of HDF5, it is probably necessary anyway to activate MPI, even without --with-mpi in the bundle.

CMakeLists.txt
piotrows: Removed.

CMakeLists.txt
piotrows: As stated earlier, dependence of hdf5.h on mpi.h. I am not sure if CXX will be needed here, but it is hard to exclude at the moment.

bundle.yml
piotrows: We are not necessarily building Fiat, so I would propose to leave it.

src/common/CMakeLists.txt
piotrows: done

PR #122, opened 2025-07-18
common: Add PP guard around OMP TARGET directives

reuterbal:
This should fix compilation with LLVM 20 by not exposing the OpenMP directives unless we have detected compiler support for OpenMP target.

Thanks to @antoine-morvan for reporting this.

### DIFF
--- a/src/common/CMakeLists.txt
+++ b/src/common/CMakeLists.txt
@@ -159,6 +159,8 @@ if( HAVE_OMP AND HAVE_OMP_TARGET )
     cloudsc_add_common_lib(
         SUFFIX omp
         TYPE ${GPU_LINK_TYPE}
+        DEFINITIONS
+            HAVE_OMP_TARGET
         LIBS
             OpenMP::OpenMP_Fortran
     )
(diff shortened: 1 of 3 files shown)


antoine-morvan:
Hello, 

Just to clarify : LLVM 20 has OpenMP target support ; just that even when you do not specify an offload target in the compiler flags, it still generates a call to the offload runtime, which will be absent at link, thus the link error.

I did not check how the `HAVE_OMP_TARGET` is initialized in the CMake files, but that not might be sufficient. I'll play with this branch and report back.

reuterbal:
Thanks @antoine-morvan ,

`HAVE_OMP_TARGET` is initialized here: https://github.com/ecmwf-ifs/dwarf-p-cloudsc/blob/develop/cmake/features/OMP.cmake
This uses a test-compile with different OpenMP target directives to determine which feature set is supported. I would expect that the link-time failure would then also occur in these tests, thus marking OpenMP target as not available.

antoine-morvan:
> Thanks @antoine-morvan ,
> 
> `HAVE_OMP_TARGET` is initialized here: https://github.com/ecmwf-ifs/dwarf-p-cloudsc/blob/develop/cmake/features/OMP.cmake This uses a test-compile with different OpenMP target directives to determine which feature set is supported. I would expect that the link-time failure would then also occur in these tests, thus marking OpenMP target as not available.

Indeed, HPCW could build CloudSC with LLVM 20 with this patch (and some hacking around a missing -lm in the linking flags, most likely related to HPCW things), and the small test terminated successfully.

mlange05 (approved):
LGTM. Many thanks for the quick fix.

PR #124, opened 2025-09-04
Improve CMake layer to facilitate compilation on Ubuntu desktops

piotrows:
This PR improves two functionalities:

1. In case of default and archaic CUDA architecture specification, code will attempt to revert to native (i.e. hardware-provided) CUDA architecture or fail at config time, demanding proper specification, e.g. with arch file.
2. It updates mix variant with the specification of the Fortran module directory (otherwise it fails to compile).
Build type tested here:
**./cloudsc-bundle build --with-mpi=ON --with-field-api=ON --with-atlas=ON --with-acc=ON --with-cuda=ON --clean --cmake="CMAKE_CUDA_COMPILER=/opt/nvidia/hpc_sdk/Linux_x86_64/2025/cuda/12.9/bin/nvcc CUDAToolkit_ROOT=/opt/nvidia/hpc_sdk/Linux_x86_64/2025/cuda/12.9"**


piotrows:
@reuterbal could you please inspect why the CI is failing? Same problems as in the other PR, but this one is only a minor bump for a pure develop.
@MichaelSt98 I think I have found a better safety valve for the CUDA version, please see if you like it.

cmake/features/CUDA.cmake
```diff
@@ -13,6 +13,20 @@ if( HAVE_CUDA )
   if( NOT DEFINED CMAKE_CUDA_ARCHITECTURES )
     set( CLOUDSC_CUDA_FLAGS "-lineinfo -maxrregcount=128" )
   else()
+    foreach(cuda_arch IN LISTS CMAKE_CUDA_ARCHITECTURES)
+       if(cuda_arch LESS 70)
+           message(WARNING "Specified CUDA architecture ${cuda_arch} is unsupported and will not compile.
```
reuterbal:
```suggestion
           ecbuild_warn("Specified CUDA architecture ${cuda_arch} is unsupported and will not compile. 
```

cmake/features/CUDA.cmake
```diff
@@ -13,6 +13,20 @@ if( HAVE_CUDA )
   if( NOT DEFINED CMAKE_CUDA_ARCHITECTURES )
     set( CLOUDSC_CUDA_FLAGS "-lineinfo -maxrregcount=128" )
   else()
+    foreach(cuda_arch IN LISTS CMAKE_CUDA_ARCHITECTURES)
+       if(cuda_arch LESS 70)
+           message(WARNING "Specified CUDA architecture ${cuda_arch} is unsupported and will not compile. 
+                            Minimum required >= 70. I am attempting to use native CUDA architecture instead.")
+           if( ${CMAKE_CUDA_ARCHITECTURES_NATIVE} LESS 70 )
+             message(FATAL_ERROR  "Native CUDA architecture ${CMAKE_CUDA_ARCHITECTURES_NATIVE} is also unsupported.
```
reuterbal:
```suggestion
             ecbuild_error("Native CUDA architecture ${CMAKE_CUDA_ARCHITECTURES_NATIVE} is also unsupported.
```

cmake/features/CUDA.cmake
```diff
@@ -13,6 +13,20 @@ if( HAVE_CUDA )
...
+                                   Please provide proper arch file or CMAKE_CUDA_ARCHITECTURES at build time or deactivate CUDA variant.")
+           else()
+             set( CMAKE_CUDA_ARCHITECTURES ${CMAKE_CUDA_ARCHITECTURES_NATIVE})
+             message(WARNING "CUDA architecture is now set to ${CMAKE_CUDA_ARCHITECTURES_NATIVE} (native version)." )
```
reuterbal:
```suggestion
             ecbuild_warn( "CUDA architecture is now set to ${CMAKE_CUDA_ARCHITECTURES_NATIVE} (native version)." )
```

reuterbal (changes requested):
Thanks for this. The missing module directory specification for mix variants was definitely a bug.
The CUDA arch logic is fine with me but I suggest to use the ecbuild logging features.

cmake/features/CUDA.cmake
piotrows: Thanks, this makes more sense. Updated now.

MichaelSt98:
I can live with that. However, only the CUDA OPT variant really requires CUDA_ARCH >= 70, all the other variants can run on older chips too.

reuterbal:
In that case it would make more sense to add this version check directly to this variant, I think

piotrows:
> In that case it would make more sense to add this version check directly to this variant, I think

You can't do that, as where it is now, the compiler options are modified. I don't think that overwriting compiler options inside a variant is a good idea. But I can add a note next to  the error message that the user can remove this restriction if needed. Anyway, sm65 cards seem rather ancient these days.

reuterbal (approved):
OK, let's leave it like this, looks good to me now.

PR #125, opened 2026-02-06
Replace FIELD_GANG with FIELD_STACK

awnawab:
The FIELD_GANG is now deprecated and removed from the FIELD_API main branch. This PR switches over the cloudsc-dwarf to use the FIELD_STACK instead. Once merged, I can rebase develop-ecwmf over main.


reuterbal (approved):
Looks good, many thanks!

Not too worried about the failing LUMI tests, this is related to the abysmall file system performance on the system: https://lumi-supercomputer.eu/lumi-service-status/information-lustre-filesystem-performance-is-very-slow/

wertysas (approved):
This is a very nice display of how easy it is to use the field stack without specifying a member map!

PR #126, opened 2026-02-13
Update Atlas to 0.45.0 to ensure compatibility with CCE

reuterbal:
That should restore LUMI hpc-ci


reuterbal:
#127 is hopefully a better fix.

PR #127, opened 2026-02-13
Add LUMI-G CCE 19 arch and use in Github Actions workflow for LUMI

reuterbal:
This is hopefully a better fix than #126


MichaelSt98 (approved):
Successfully able to reproduce.

